PR #3, opened 2020-05-03
Make recipe arguments named, with good error messaging

jpslav:
This will allow us to pass other named (optional) arguments in the future


jpslav:
Wonder if it should still allow a non-keyword argument as first argument to have whatever name the recipe writer wants, and then some other named arguments that can be filled in with other utility things like `counters`

jpslav:
nah

PR #41, opened 2020-12-22
'safe' rubocop fixes

jpslav:
Automatic Rubocop "safe" fixes (`rubocop -a`), plus some Rubocop exceptions and default configuration changes.

Later when we have more specs we can try the "unsafe" autocorrections with `rubocop -A`.


jpslav:
Specs will get added in later PRs, that this PR touched every file for linting fixes shouldn't count against it for codecov.

Issue #46, opened 2021-01-15
Write a spec for the `Kitchen::Counter` class


jpslav:
Use `RSpec` (standard Ruby testing framework) to write a set of tests for the `Kitchen::Counter` class.  This class keeps track of an integer count and has 3 or 4 methods.

PR #74, opened 2021-01-19
Add specs for counter class

kswanson33:
#46

### DIFF
--- a/spec/counter_spec.rb
+++ b/spec/counter_spec.rb
@@ -1,0 +1,45 @@
+require 'spec_helper'
+
+RSpec.describe Kitchen::Counter do
+  describe ".initialize" do
+    it 'creates a counter' do
+      expect(Kitchen::Counter.new).to be_truthy
+    end
+  end
+  describe ".increment" do
+    counter = Kitchen::Counter.new
+    it 'increments by 1 given no args' do
+      val = counter.instance_variable_get(:@value)
+      counter.increment
+      expect(counter.instance_variable_get(:@value)).to eq(val + 1)
+    end
+    it 'increments by 2' do
+      val = counter.instance_variable_get(:@value)
+      counter.increment(by: 2)
+      expect(counter.instance_variable_get(:@value)).to eq(val + 2)
+    end
+    it 'increments by -5' do
+      val = counter.instance_variable_get(:@value)
+      counter.increment(by: -5)
+      expect(counter.instance_variable_get(:@value)).to eq(val -5)
+    end
+  end
+  describe ".get" do
+    it 'gets the counter value' do
+      counter = Kitchen::Counter.new
+      expect(counter.get).to eq(counter.instance_variable_get(:@value))
+    end
+  end
+  describe ".reset" do
+    counter = Kitchen::Counter.new
+    counter.increment
+    it 'resets to zero given no args' do
+      counter.reset
+      expect(counter.instance_variable_get(:@value)).to eq(0)
+    end
+    it 'resets to a given value' do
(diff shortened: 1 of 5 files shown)


spec/counter_spec.rb
```diff
@@ -0,0 +1,45 @@
+require 'spec_helper'
+
+RSpec.describe Kitchen::Counter do
+  describe '.initialize' do
+    it 'creates a counter' do
+      expect(Kitchen::Counter.new).to be_truthy
+    end
+  end
+  describe '.increment' do
+    counter = Kitchen::Counter.new
```
jpslav, line 10:
```ruby
RSpec.describe Kitchen::Counter do
  let(:instance) { described_class.new }
```

You can then use `instance` instead of saying `counter = Kitchen::Counter.new` below

spec/counter_spec.rb
```diff
@@ -0,0 +1,45 @@
...
+    it 'increments by -5' do
+      val = counter.instance_variable_get(:@value)
+      counter.increment(by: -5)
+      expect(counter.instance_variable_get(:@value)).to eq(val - 5)
+    end
+  end
+  describe '.get' do
+    it 'gets the counter value' do
+      counter = Kitchen::Counter.new
+      expect(counter.get).to eq(counter.instance_variable_get(:@value))
```
jpslav, line 30: I think it is ok to only test the public interface and not look at the internal value.  We know that `get` just returns `@value`, and our other code will make sure that `get` gives good results.

spec/counter_spec.rb
```diff
@@ -0,0 +1,45 @@
...
+    it 'creates a counter' do
+      expect(Kitchen::Counter.new).to be_truthy
+    end
+  end
+  describe '.increment' do
+    counter = Kitchen::Counter.new
+    it 'increments by 1 given no args' do
+      val = counter.instance_variable_get(:@value)
+      counter.increment
+      expect(counter.instance_variable_get(:@value)).to eq(val + 1)
```
jpslav, line 14: Way to go on getting to the member variable -- advance! But per below, I think it is ok to just test based on `get`.

spec/counter_spec.rb
```diff
@@ -0,0 +1,45 @@
...
+    end
+    it 'increments by 2' do
+      val = counter.instance_variable_get(:@value)
+      counter.increment(by: 2)
+      expect(counter.instance_variable_get(:@value)).to eq(val + 2)
```
jpslav, line 19: This logic is totally right but there's a more compact way to write it

```ruby
expect { counter.increment(by: 2) }.to change { counter.get }.from(0).to(2)
```

See https://relishapp.com/rspec/rspec-expectations/docs/built-in-matchers/change-matcher

jpslav (changes requested):

spec/counter_spec.rb
jpslav: Note that this passes blocks around (see the curly braces) instead of passing arguments using parentheses.  The code in the blocks is passed into `expect` so it can run the code when it is ready.  Internally it runs `counter.get` first to get the initial value, then it runs `counter.increment(by: 2)`, then it runs `counter.get` again

spec/counter_spec.rb
kswanson33: So to clarify, you mean I should just use `get` in my other specs instead of `instance_variable_get`, and take out this spec?

kswanson33: Hah, just scrolled down and saw your other comment. Disregard!

jpslav (approved):

PR #76, opened 2021-01-19
spec for ancestor

brittany-johnson:


lib/kitchen/ancestor.rb
```diff
@@ -38,7 +38,11 @@ def increment_descendant_count(descendant_type)
     # @param by [Integer] the amount by which to decrement
     #
     def decrement_descendant_count(descendant_type, by: 1)
-      @descendant_counts[descendant_type.to_sym] = get_descendant_count(descendant_type) - by
+      if @descendant_counts.has_key?(descendant_type.to_sym)
+        @descendant_counts[descendant_type.to_sym] = get_descendant_count(descendant_type) - by
+      else
+        raise "An element cannot have negative descendants"
```
jpslav, line 44: Great idea to have this protection!  However note that it does not fully protect against negative counts.  If the count already existed at `1` and you had `by: -2` it'd go negative.  You could add that spec first and then fix here.

jpslav: Also modify the YARD documentation above this method that this method can raise an error (there are other examples of this)

spec/ancestor_spec.rb
```diff
@@ -0,0 +1,47 @@
...
+    expect(ancestor.type).to eq element.short_type
+  end
+
+  it 'makes the element available' do
+    expect(ancestor.element).to eq element
+  end
+
+  describe '#increment_descendant_count' do
+    it 'adds 1 to the descendant count for the given type' do
+      expect { ancestor.increment_descendant_count(descendant.short_type) }.to change {ancestor.get_descendant_count(descendant.short_type)}.by(1)
```
jpslav, line 20: rubocop doesn't complain about line length b/c we have turned off that check in specs -- because we can often have long lines.  but in this case it is pretty long and we can organize it into more shorter lines:

```ruby
expect { 
  ancestor.increment_descendant_count(descendant.short_type) 
}.to change {
  ancestor.get_descendant_count(descendant.short_type)
}.by(1)
```

spec/ancestor_spec.rb
```diff
@@ -0,0 +1,47 @@
...
+
+  describe '#decrement_descendant_count' do
+    it 'decreases the descendant count for the given type by 1' do
+      ancestor.increment_descendant_count(descendant.short_type)
+      expect { ancestor.decrement_descendant_count(descendant.short_type) }.to change {ancestor.get_descendant_count(descendant.short_type)}.by(-1)
+    end
+    it 'decreases the descendant count for the given type by -5' do
+      expect { ancestor.decrement_descendant_count(descendant.short_type, by: -5) }.to raise_error("An element cannot have negative descendants")
+    end
+  end
```
jpslav, line 32: per above add the test where you increment by 1 then decrement by 2

spec/ancestor_spec.rb
```diff
@@ -0,0 +1,47 @@
+require 'spec_helper'
+require 'ostruct'
+
+RSpec.describe Kitchen::Ancestor do
+
+  let(:element) { OpenStruct.new(short_type: 'page') }
+  let(:descendant) { OpenStruct.new(short_type: 'note')}
```
jpslav, line 7: you don't actually use `descendant` so you could replace this with 

```ruby
let(:descendent_type) { 'note' }
```

spec/ancestor_spec.rb
```diff
@@ -0,0 +1,47 @@
...
+  describe '#get_descendant_count' do
+    it 'returns the descendant count for the given type' do
+      ancestor.increment_descendant_count(descendant.short_type)
+      expect(ancestor.get_descendant_count(descendant.short_type)).to eq 1
+    end
+  end
+
+  describe '#clone' do
+    it 'makes a new ancestor around the same element with new counts' do
+      @ancestor_clone = ancestor.clone()
```
jpslav, line 43: * You don't need the `()` when there are no arguments; it is more Ruby-like to just say `ancestor.clone`
* Don't need the `@` -- use that for instance variables in classes where you need to use the variable across methods.
* I recommend incrementing the descendant count on the ancestor before you clone it, that way you'll know that the `0` you're seeing in the clone is a true reset of the counts and not just what also happens to be in the original.

jpslav (changes requested):
Looking good!

lib/kitchen/ancestor.rb
```diff
@@ -36,9 +36,15 @@ def increment_descendant_count(descendant_type)
     #
     # @param descendant_type [String, Symbol] the descendent's type
     # @param by [Integer] the amount by which to decrement
+    # @raise [RangeError] if descendant count is a negative number
```
brittany-johnson, line 39: I'm not sure if RangeError was the right choice here. 
I was also considering Standard and Argument?

lib/kitchen/ancestor.rb
```diff
@@ -36,9 +36,15 @@ def increment_descendant_count(descendant_type)
...
     #
     def decrement_descendant_count(descendant_type, by: 1)
-      @descendant_counts[descendant_type.to_sym] = get_descendant_count(descendant_type) - by
+      if (@descendant_counts.has_key?(descendant_type.to_sym)) &&
```
brittany-johnson, line 42: This conditional is super long, not sure if I indented properly

lib/kitchen/ancestor.rb
jpslav: that's right b/c that's what you're raising.

spec/ancestor_spec.rb
jpslav: Please add the spec where you decrement without having already incremented.

jpslav (changes requested):

jpslav (approved):

Issue #51, opened 2021-01-15
Complete the spec for `Kitchen::SearchHistory`


jpslav:
Finish the spec for `Kitchen::SearchHistory`.  There's already a spec file for it with two tests, but there are more methods to test.  

If you look at test coverage, you'll see that [this class is mostly covered already](https://codecov.io/gh/openstax/kitchen/src/master/lib/kitchen/ancestor.rb) -- that's because other higher-level specs test code that makes a lot of use of search history. Nevertheless, for such a fundamental class we should have a full spec dedicated to it, which you'll add via this issue.

PR #77, opened 2021-01-20
Add additional specs for search_history

kswanson33:
#51

### DIFF
--- a/spec/search_history_spec.rb
+++ b/spec/search_history_spec.rb
@@ -1,15 +1,49 @@
 require 'spec_helper'
 
 RSpec.describe Kitchen::SearchHistory do
+  context 'init' do
+    it 'creates instance variables' do
+      sh = described_class.new('foo', 'bar')
+      expect(sh.upstream).to eq('foo')
+      expect(sh.latest).to eq('bar')
+    end
+  end
+
   context 'empty' do
     it 'converts to an empty array' do
       expect(described_class.empty.to_a).to eq []
     end
   end
 
+  context '#add' do
+    it 'adds nil if given nil value' do
+      expect(described_class.empty.add(nil).to_a).to eq [nil]
+    end
+  end
+
   context '#to_s' do
     it 'works' do
       expect(described_class.empty.add(nil).add("foo").add([".blah",".bar"]).to_s).to eq "[?] [foo] [.blah, .bar]"
+    end
+  end
+
+  context '#to_a' do
+    it 'returns [latest] if no upstream' do
+      expect(described_class.empty.add('foo').to_a).to eq(['foo'])
+    end
+
+    it 'returns array with upstream and latest' do
+      expect(described_class.empty.add('foo').add(["blah","bar"]).add(['baz']).to_a).to eq(['foo','blah, bar','baz'])
+    end
+  end
+
+  context "#empty?" do
+    it 'returns true if empty' do
+      expect(described_class.empty.empty?).to eq(true)
+    end
+
+    it 'returns false if not empty' do
+      expect(described_class.empty.add('foo').empty?). to eq(false)
     end
   end
 end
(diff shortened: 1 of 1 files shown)


spec/search_history_spec.rb
```diff
@@ -1,15 +1,49 @@
 require 'spec_helper'
 
 RSpec.describe Kitchen::SearchHistory do
+  context 'init' do
+    it 'creates instance variables' do
+      sh = described_class.new('foo', 'bar')
+      expect(sh.upstream).to eq('foo')
+      expect(sh.latest).to eq('bar')
```
jpslav, line 8: way to go with the fake arguments! 👏

spec/search_history_spec.rb
```diff
@@ -1,15 +1,49 @@
...
   context 'empty' do
     it 'converts to an empty array' do
       expect(described_class.empty.to_a).to eq []
     end
   end
 
+  context '#add' do
+    it 'adds nil if given nil value' do
+      expect(described_class.empty.add(nil).to_a).to eq [nil]
+    end
```
jpslav, line 21: Nevermind ~Looking at the `[css_or_xpath].join(', ')` part of `#add` I'm confused why I did that.  You might look through the codebase and see if we ever pass in an array as the `css_or_xpath` -- that's the only reason I can think of that we'd need a `join` but at the same time if `css_or_xpath` is an array, then we have nested arrays which is also confusing.~

jpslav: I see that I used `to_a` when testing, but there might be a better way.  Here you could maybe say

```ruby
expect(described_class.empty.add(nil).latest).to be_nil
```

spec/search_history_spec.rb
```diff
@@ -1,15 +1,49 @@
...
+
+    it 'returns array with upstream and latest' do
+      expect(described_class.empty.add('foo').add(["blah","bar"]).add(['baz']).to_a).to eq(['foo','blah, bar','baz'])
+    end
+  end
+
+  context "#empty?" do
+    it 'returns true if empty' do
+      expect(described_class.empty.empty?).to eq(true)
```
jpslav, line 42:
```ruby
expect(described_class.empty).to be_empty
```

https://relishapp.com/rspec/rspec-expectations/v/3-10/docs/built-in-matchers/predicate-matchers

spec/search_history_spec.rb
```diff
@@ -1,15 +1,49 @@
...
+    end
+
+    it 'returns false if not empty' do
+      expect(described_class.empty.add('foo').empty?). to eq(false)
```
jpslav, line 46:
```ruby
expect(described_class.empty).not_to be_empty
```

spec/search_history_spec.rb
```diff
@@ -1,15 +1,49 @@
...
     end
   end
+
+  context '#to_a' do
+    it 'returns [latest] if no upstream' do
+      expect(described_class.empty.add('foo').to_a).to eq(['foo'])
+    end
+
+    it 'returns array with upstream and latest' do
+      expect(described_class.empty.add('foo').add(["blah","bar"]).add(['baz']).to_a).to eq(['foo','blah, bar','baz'])
```
jpslav, line 36: rubocop should be complaining about the double quoted strings, changing them to single quoted.

jpslav (changes requested):
Looking good!

spec/search_history_spec.rb
kswanson33: Just to check that I'm understanding correctly -- the predicate matcher `be_empty` will call the `#empty?` method defined on `SearchHistory` under the hood?

jpslav: yep. it'll split `be_empty` to get `empty` and add the `?` for you.

jpslav (approved):

Issue #49, opened 2021-01-15
Write the spec for the `Kitchen::Clipboard` class


jpslav:
Use RSpec (standard Ruby testing framework) to write a set of tests for the `Kitchen::Clipboard` class. This class is a glorified array.  In Kitchen we copy to and paste items from clipboards as we move content around.

There's already an existing spec file but it logically empty :-)  Note that the one spec in it is disabled.  Instead of saying `it 'returns self to enable chaining'` it starts with `xit` which is rspec-speak for "disable this test".  So if you run it and it isn't working make sure to take off the `x` :-)   Also note that you'll need to add more specs than are sketched out in the current spec file.

If you look at test coverage, you'll see that [this class is mostly covered already](https://codecov.io/gh/openstax/kitchen/src/master/lib/kitchen/clipboard.rb) -- that's because other higher-level specs test code that makes a lot of use of clipboards.  Nevertheless, for such a fundamental class we should have a full spec dedicated to it, which you'll add via this issue.

PR #78, opened 2021-01-21
Add specs for coverage clipboard class

KarinaMendez2:
#49

### DIFF
--- a/spec/clipboard_spec.rb
+++ b/spec/clipboard_spec.rb
@@ -1,11 +1,85 @@
 require 'spec_helper'
 
 RSpec.describe Kitchen::Clipboard do
+  let(:my_clipboard) { described_class.new }
+  let(:element_1) do
+    new_element(
+      <<~HTML
+        <div id='e1'>
+          <span>Element 1</span>
+          <span>ABC</span>
+          <span>Zebra</span>
+          <span>B</span>
+        </div>
+      HTML
+    )
+  end
+
+  let(:element_2) do
+    new_element(
+      <<~HTML
+        <h1 id='e2'>Title</h1>
+      HTML
+    )
+  end
+
+  context '#add' do
+    it 'throws error when no item was given' do
+      expect{my_clipboard.add}.to raise_error(ArgumentError)
+    end
+
+    it 'returns array with added item' do
+      my_clipboard.add(1)
+      expect(my_clipboard.add(2)).to eq([1, 2])
+    end
+  end
+
+  context '#clear' do
+    it 'clears clipboard contents' do
+      expect(my_clipboard.add(1)).to eq([1])
+      expect(my_clipboard.clear).to be {}
+    end
+  end
+
   context '#paste' do
+    it 'returns a concatenation of the pasting' do
+      element_2.cut(to: my_clipboard)
+      element_1.cut(to: my_clipboard)
+      expect(my_clipboard.paste).to match(/[e1][e2]/)
+    end
+
+    it 'returns empty object when theres nothing to paste' do
+      expect(my_clipboard.paste).to be {}
+    end
   end
 
   context '#each' do
-    xit 'returns self to enable chaining' do
+    it 'returns self when the clipboard has contents' do
+      element_1.cut(to: my_clipboard)
+      expect(my_clipboard.each{}).to be my_clipboard
+    end
+
+    it 'returns self when the clipboard is empty' do
+      expect(my_clipboard.each).to be my_clipboard
+    end
+
+    it 'handles clipboard not empty and no block given' do
+      element_1.cut(to: my_clipboard)
+      expect{my_clipboard.each}.to raise_error(NoMethodError)
+    end
+  end
+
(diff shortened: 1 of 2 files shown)


spec/clipboard_spec.rb
```diff
@@ -1,11 +1,85 @@
...
+    it 'handles clipboard not empty and no block given' do
+      element_1.cut(to: my_clipboard)
+      expect{my_clipboard.each}.to raise_error(NoMethodError)
+    end
+  end
+
+  context '#sort_by' do
+    it 'returns empty object when no block given' do
+      element_1.cut(to: my_clipboard)
+      expect(my_clipboard.sort_by!).to be {}
```
jpslav, line 75: Ditto here on `{}` being read as a block being passed to the `be` method, not being read as a hash.

[`Array#sort_by!`](https://apidock.com/ruby/v2_5_5/Array/sort_by%21) returns an Enumerator is no block is given so I think our test should look for the same

```ruby
expect(my_clipboard.sort_by!).to be_a Enumerator
```

`sort_by!`

spec/clipboard_spec.rb
```diff
@@ -1,11 +1,85 @@
...
+    end
+
+    it 'returns sorted object when block given' do
+      element_1.search('span').cut(to: my_clipboard)
+      expect(
+        my_clipboard.sort_by!{|element| element.text}.to_s
+      ).to eq("[<span>ABC</span>, <span>B</span>, <span>Element 1</span>, <span>Zebra</span>]")
```
jpslav, line 82: You can use `fake_element` here again and sort by the value of the paste method:  `my_clipboard.sort_by!(&:paste)`

jpslav (changes requested):
Looking good!

spec/clipboard_spec.rb
```diff
@@ -1,85 +1,76 @@
...
 
-    it 'handles clipboard not empty and no block given' do
-      element_1.cut(to: my_clipboard)
-      expect{my_clipboard.each}.to raise_error(NoMethodError)
+      it 'when clipboard is not empty and no block is given' do
+        my_clipboard.add(2)
+        expect(my_clipboard.each).to be my_clipboard
+      end
     end
   end
```
jpslav, line 59: Need a spec to make sure that each works.  This is a tricky one so I will share the answer and other devs should take a look too.

One way is more manual.  You can create a lambda which is a callable bit of code (here there is actually no code in the `{}` block).  And then you can expect the block to be called with the items on the clipboard.  And then you add the items and call each, passing in the block (`&a_block` does that).

```ruby
it 'iterates on elements ' do
  a_block = ->(item) {}
  expect(a_block).to receive(:call).with("howdy")
  expect(a_block).to receive(:call).with("y'all")
  my_clipboard.add("howdy").add("y'all").each(&a_block)
end
```

It turns out that [rspec has a built-in way to test blocks](https://relishapp.com/rspec/rspec-expectations/docs/built-in-matchers/yield-matchers#yield-with-args-matcher) with less code:

```ruby
it 'iterates on elements' do
  my_clipboard.add("howdy").add("y'all")
  expect{ |block| my_clipboard.each(&block) }.to yield_successive_args("howdy", "y'all")
end
```

spec/clipboard_spec.rb
```diff
@@ -1,85 +1,76 @@
...
 
   context '#sort_by' do
     it 'returns empty object when no block given' do
-      element_1.cut(to: my_clipboard)
-      expect(my_clipboard.sort_by!).to be {}
+      my_clipboard.add(fake_element(2))
```
jpslav, line 63: don't think you need this line

spec/clipboard_spec.rb
```diff
@@ -1,85 +1,76 @@
...
 
     it 'returns sorted object when block given' do
-      element_1.search('span').cut(to: my_clipboard)
+      my_clipboard.add(fake_element('Zebra'))
+      my_clipboard.add(fake_element('ABC'))
+      my_clipboard.add(fake_element('Element 1'))
       expect(
-        my_clipboard.sort_by!{|element| element.text}.to_s
-      ).to eq("[<span>ABC</span>, <span>B</span>, <span>Element 1</span>, <span>Zebra</span>]")
+        my_clipboard.sort_by!(&:paste)[0].paste
```
jpslav, line 72: `Clipboard` has a `paste` method itself (as do the fake elements), so if you take out the `[0]` you can test that the other elements are sorted too `'ABCElement 1Zebra'`.

jpslav (changes requested):

spec/clipboard_spec.rb
KarinaMendez2: Here I was actually getting `undefined method `paste' for #<Array:'...'>` so I printed the result to the console, it comes out like: `#<OpenStruct paste="ABC"> #<OpenStruct paste="Element">#<OpenStruct paste="Zebra">`

jpslav (approved):

Issue #47, opened 2021-01-15
Create and test `BakeChapterTitle` direction


jpslav:
Move [code from the Chemistry 2e recipe](https://github.com/openstax/recipes/blob/main/books/chemistry2e/bake#L39-L49) into a new `BakeChapterTitle` direction.

Do this in the style of [`BakeFootnotes`](https://github.com/openstax/kitchen/tree/master/lib/kitchen/directions/bake_footnotes), i.e.

1. Make a `lib/kitchen/directions/bake_chapter_title` directory
2. Add `main.rb` and `v1.rb` files like in `bake_footnotes`
3. Add `spec/directions/bake_chapter_title/v1_spec.rb` to contain the test, see the footnotes version of this as a guide.

PR #84, opened 2021-01-22
Add spec for bake_chapter_title

kswanson33:
#47 
Copies implementation for BakeChapterTitle from [recipes](https://github.com/openstax/recipes/blob/main/books/chemistry2e/bake#L39-L49).

### DIFF
--- a/lib/kitchen/directions/bake_chapter_title/main.rb
+++ b/lib/kitchen/directions/bake_chapter_title/main.rb
@@ -1,0 +1,9 @@
+module Kitchen
+  module Directions
+    module BakeChapterTitle
+      def self.v1(book:)
+        V1.new.bake(book: book)
+      end
+    end
+  end
+end
--- a/lib/kitchen/directions/bake_chapter_title/v1.rb
+++ b/lib/kitchen/directions/bake_chapter_title/v1.rb
@@ -1,0 +1,23 @@
+module Kitchen::Directions::BakeChapterTitle
+  class V1
+    def bake(book:)
+      book.chapters.each do |chapter|
+        fix_up_chapter_title(chapter: chapter)
+      end
+    end
+
+    def fix_up_chapter_title(chapter:)
+      # Fix up chapter titles
+      heading = chapter.at('h1[2]')
+      heading[:id] = "chapTitle#{chapter.count_in(:book)}"
+      heading.replace_children(with:
+        <<~HTML
+          <span class="os-part-text">Chapter </span>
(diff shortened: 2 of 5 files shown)


spec/.rubocop.yml
```diff
@@ -23,3 +23,6 @@ Style/SymbolProc:
 
 RSpec/FilePath:
   Enabled: false
+
+RSpec/ExampleLength:
+  Enabled: false
```
kswanson33, line 28: This addition gets rid of the 5-line maximum for a spec example -- lmk if we want to do something different.

jpslav: Yeah I think this is fine.  in non rspec-land, there are some settings like

```ruby
Metrics/MethodLength:
  Max: 50
  CountAsOne: ['heredoc']
```

that would count your big heredoc string as one line, but I don't think the rspec cop has this feature

lib/kitchen/directions/bake_chapter_title/main.rb
```diff
@@ -0,0 +1,9 @@
+module Kitchen
+  module Directions
+    module BakeChapterTitle
+      def self.v1(book:)
+        V1.new.bake(book: book)
```
jpslav, line 5: You can write a simple test for this line:

```ruby
RSpec.describe Kitchen::Directions::BakeChapterTitle do
  it 'calls v1' do
    expect_any_instance_of(Kitchen::Directions::BakeChapterTitle::V1).to receive(:bake).with(book: "blah")
    described_class.v1(book: "blah")
  end
end
```

Later in this file we'd add `it 'calls v2'` etc

lib/kitchen/directions/bake_chapter_title/v1.rb
```diff
@@ -0,0 +1,23 @@
+module Kitchen::Directions::BakeChapterTitle
+  class V1
+    def bake(book:)
+      book.chapters.each do |chapter|
+        fix_up_chapter_title(chapter: chapter)
+      end
+    end
+
+    def fix_up_chapter_title(chapter:)
+      # Fix up chapter titles
```
jpslav, line 10: can probably ditch the comment as the method name tells us what we need to know

lib/kitchen/directions/bake_chapter_title/main.rb
jpslav: I'm sure this is missing for the other existing directions specs too

jpslav (changes requested):

jpslav (approved):

PR #85, opened 2021-01-25
Creates config_spec.rb

LostMojo:
checks that the class can be initialized, read its selectors, and that Config.new_from_file isn't implemented.

### DIFF
--- a/spec/config_spec.rb
+++ b/spec/config_spec.rb
@@ -15,10 +15,9 @@ RSpec.describe Kitchen::Config do
     end
   end
 
-  #NYI
   describe 'self.new_from_file' do
     it "isn't implemented yet" do
-      expect{ described_class.new_from_file('bleh') }.to raise_error('NYI')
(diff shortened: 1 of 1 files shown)


spec/config_spec.rb
```diff
@@ -0,0 +1,25 @@
...
+      expect(foobar.selectors).to eq('foo')
+    end
+  end
+
+  #NYI
+  describe 'self.new_from_file' do
+    it "isn't implemented yet" do
+      expect{ described_class.new_from_file('bleh') }.to raise_error('NYI')
+    end
+  end
```
jpslav, line 23: There are two rubocop errors down around here 
image

jpslav (changes requested):
Looks good, minor cleanup needed.

spec/config_spec.rb
LostMojo: gottem!

jpslav (approved):

PR #86, opened 2021-01-25
Create string_spec.rb

LostMojo:


spec/string_spec.rb
```diff
@@ -0,0 +1,11 @@
+require 'spec_helper'
+
+RSpec.describe String do
+
+  describe '#uncapitalize' do
+    it 'downcases the first letter of self, returning a new string' do
+      expect('Tomato'.uncapitalize).to eq('tomato')
```
jpslav, line 7: Can you modify the spec to have some other capitalized letters in the word to ensure that it is only the first letter that is being downcased?

jpslav (changes requested):

spec/string_spec.rb
LostMojo: done!

jpslav (approved):

Issue #48, opened 2021-01-15
Write a spec for the `Kitchen::Pantry` class


jpslav:
Use RSpec (standard Ruby testing framework) to write a set of tests for the `Kitchen::Pantry` class. This class wraps a hash (like a dictionary in Python) and has a few methods for storing things and getting things out of the pantry.

PR #87, opened 2021-01-26
Create an initial spec for Kitchen::Pantry

Stackblocks:
Tests for `.instance`, `.store`, `.get`, `.get!`, `.each` and `.size` methods in the Pantry class.

Re: #48

### DIFF
--- a/spec/pantry_spec.rb
+++ b/spec/pantry_spec.rb
@@ -1,0 +1,30 @@
+require 'spec_helper'
+
+RSpec.describe Kitchen::Pantry do
+  let (:instance) {described_class.new}
+
+  it 'creates a hash' do
+      expect(described_class.new).to be_truthy
+  end
+
+  it 'adds an item to the Pantry with the provided label' do
+    instance.store 'bar', label: 'foo'
+    expect { instance }.not_to be_empty
+  end
+
+  it 'gets an item from the Pantry' do
+    expect { instance.get('foo') }.to eq('bar')
+  end
+
+  it 'gets an item from the pantry, raising if not present' do
+    expect { instance.get!('thud')}.to raise_error(Kitchen::RecipeError)
+  end
+
+  it 'iterate over the pantry items' do
+    expect { instance.each(&block) }.to include(TK)
+  end
+
+  it 'returns the number of items in the pantry' do
+    expect { instance.size }.to eq(1)
+  end
+end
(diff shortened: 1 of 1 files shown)


spec/pantry_spec.rb
```diff
@@ -0,0 +1,36 @@
+require 'spec_helper'
+
+RSpec.describe Kitchen::Pantry do
+  let(:instance) { described_class.new }
+
+  it 'creates a hash' do
```
jpslav, line 6: the constructor creates a pantry object, not a hash, so need to change this spec name

spec/pantry_spec.rb
```diff
@@ -0,0 +1,36 @@
...
+    expect(instance).to be_truthy
+  end
+
+  context 'when the pantry has contents' do
+    before { instance.store 'bar', label: 'foo' }
+
+    it 'adds an item to the pantry with the provided label' do
```
jpslav, line 13: this spec isn't adding an item so need to change the name

spec/pantry_spec.rb
```diff
@@ -0,0 +1,36 @@
...
+
+    it 'gets an item from the pantry' do
+      expect(instance.get(:foo)).to eq('bar')
+    end
+
+    it 'raises if requested item is not in pantry' do
+      expect{ instance.get!('thud') }.to raise_error(
+        Kitchen::RecipeError, "There is no pantry item labeled 'thud'"
+      )
+    end
```
jpslav, line 25: Would be good to also test the behavior of `get` (not `get!`) when the item isn't present -- it should return `nil`

So you might break it down like 

```ruby
context '#get' do
  it 'returns a present item with either a string or symbol label' do
    expect(instance.get(:foo)).to eq('bar')
    expect(instance.get('foo')).to eq('bar')
  end

  it 'returns nil when an item is not present' do
    expect(instance.get(:blah)).to be_nil
  end
end

context '#get!' do
  it 'returns a present item with either a string or symbol label' do
    expect(instance.get!(:foo)).to eq('bar')
    expect(instance.get!('foo')).to eq('bar')
  end

  it 'raises if the item is not present' do
    # your existing raise spec here
  end
end
```

jpslav (changes requested):

spec/pantry_spec.rb
Stackblocks: Good point about having positive and negative path tests for both. I've incorporated this and reworked the layout to match 👍

jpslav (approved):

PR #88, opened 2021-01-26
adds Utils specs

LostMojo:


spec/utils_spec.rb
```diff
@@ -0,0 +1,10 @@
+require 'spec_helper'
+
+RSpec.describe Kitchen::Utils do
+  describe '.search_path_to_type(search_path' do
```
jpslav, line 4: did you want the `(search_path` at the end?

LostMojo: whoops! no

jpslav (approved):

PR #89, opened 2021-01-26
Specs bake appendix

kswanson33:
#56 

(not sure why prev changes are showing up in this branch 😕 )

### DIFF
--- a/lib/kitchen/directions/bake_appendix.rb
+++ b/lib/kitchen/directions/bake_appendix.rb
@@ -18,6 +18,7 @@ module Kitchen
           title = section.first!("[data-type='title']")
           title.name = "h#{section['data-depth'].to_i + 1}"
         end
+        page
       end
     end
   end
--- a/spec/directions/bake_appendix_spec.rb
+++ b/spec/directions/bake_appendix_spec.rb
@@ -1,0 +1,43 @@
+require 'spec_helper'
+
+RSpec.describe Kitchen::Directions::BakeAppendix do
+  let(:doc) do
+    Kitchen::BookDocument.new(document: Nokogiri::XML(<<~HTML
+      <html>
+      <body>
+        <div data-type="chapter">
+          <div data-type="page">
+            <div data-type="document-title">zzzzzzz</div>
+            <section>
+              <div data-type="title">hello</div>
+              <div>world</div>
+            </section>
+          </div>
+        </div>
+      </body>
+      </html>
+    HTML
+    ))
+  end
+
+  it 'works' do
+    page = doc.book.chapters.first.pages.first
+    described_class.v1(page: page, number: 3)
+    expect(page).to match_normalized_html(
+      <<~HTML
+        <div data-type="page">
+          <h1 data-type="document-title">
+            <span class="os-part-text">Appendix </span>
+            <span class="os-number">3</span>
+            <span class="os-divider"> </span>
+            <span class="os-text" data-type="" itemprop="">zzzzzzz</span>
+          </h1>
+          <section>
+            <h1 data-type="title">hello</h1>
+            <div>world</div>
+          </section>
+        </div>
+      HTML
+    )
+  end
+end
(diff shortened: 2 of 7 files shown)


spec/directions/bake_appendix_spec.rb
```diff
@@ -0,0 +1,43 @@
...
+            <section>
+              <div data-type="title">hello</div>
+              <div>world</div>
+            </section>
+          </div>
+        </div>
+      </body>
+      </html>
+    HTML
+    ))
```
kswanson33, line 20: Is there a way to create a `PageElement` directly (i.e. without creating a `BookDocument` first)?

jpslav: You can use one of the [helper methods here](https://github.com/openstax/kitchen/blob/master/spec/helpers/factory_helpers.rb) or you can add a new `page_element` helper method in this file e.g.

```ruby
def page_element(page_children_html)
  book_containing(html: "<div data-type='page'>#{page_children_html}</div>").pages.first
end
```

spec/directions/bake_appendix_spec.rb
```diff
@@ -0,0 +1,43 @@
...
+  let(:doc) do
+    Kitchen::BookDocument.new(document: Nokogiri::XML(<<~HTML
+      <html>
+      <body>
+        <div data-type="chapter">
+          <div data-type="page">
+            <div data-type="document-title">zzzzzzz</div>
+            <section>
+              <div data-type="title">hello</div>
+              <div>world</div>
```
jpslav, line 13: Can you add a nested `<section>` tag to see that its title gets converted to `h2`?

jpslav (changes requested):

lib/kitchen/directions/bake_appendix.rb
```diff
@@ -16,8 +16,9 @@ def self.v1(page:, number:)
         # Make a section with data-depth of X have a header level of X+1
         page.search('section').each do |section|
           title = section.first!("[data-type='title']")
-          title.name = "h#{section['data-depth'].to_i + 1}"
+          title.name = "h#{section['data-depth'].to_i + 1}"  # doesn't work
```
kswanson33, line 19: @jpslav Good catch asking for me to test the section header -- tests are now failing because `section['data-depth']` isn't retrieving the right value. I took a stab at fixing this myself, but couldn't hack it.

My investigation: 
- `section['data-depth']` is empty, and `section['data-depth'].to_i` returns `0` no matter what the depth is. 
- The syntax of this implies that `section` is a hash with key `data-depth`, but `section.class` is `ElementEnumerator`
- The Ruby hash methods (`key`, `include`?) don't exist on `section`. 

Any advice?

jpslav: The `section` tag in your test needs to have a `data-depth` attribute, e.g. `data-depth=1`.  Here's what one looks like in real assembled book XHTML:

```xml
<section data-depth="1" id="auto_m68662_eip-260">
```

`section['blah']` is legit, the `[]` method on the `Kitchen::Element` class looks up the value for an attribute on the element.  Since the attribute is missing, `section['data-depth']` is `nil`, and `nil.to_i` is `0` for better or worse.

jpslav: does kind of beg the question of what to do if the `data-depth` tag isn't there...

lib/kitchen/directions/bake_appendix.rb
```diff
@@ -18,6 +18,7 @@ def self.v1(page:, number:)
           title = section.first!("[data-type='title']")
           title.name = "h#{section['data-depth'].to_i + 1}"
         end
+        page
```
jpslav, line 21: what's the thinking on this return value?

kswanson33: I was thinking about return types and how it didn’t make much sense for this function to return what it previously did, which was an integer (I believe?) — I can take out this change, or change to nil or a different value — what do you think? (Does it matter?)

kswanson33: I've reverted this

jpslav (approved):

Issue #52, opened 2021-01-15
Write spec for `ElementBase`


jpslav:
The `ElementBase` class is the parent class of all element classes.  It has a fair bit to it, so when writing the spec it is good to be organized.

When looking at any particular method in this class, particularly those with input arguments, there will be a number of different cases to test.  I recommend organizing the spec where each method has its own `context` block (`context do ... end`) within which there are a number of spec examples (`it 'blah blah' ...`):

```ruby
RSpec.describe Kitchen::ElementBase do
  context '#prepend' do
    it 'prepends a child' do
      ...
    end

    it 'prepends a sibling' do
      ...
    end 

    it 'only allows you to prepend a child or a sibling' do
      ...
    end
  
    it 'requires either a child or a sibling' do
      ...
    end 

    # etc
  end

  context '#append' do
    ...
  end
end
```

When you look at test coverage you'll see most of this class is already covered by higher level specs, but still good to have low-level spec coverage on this essential class.

PR #92, opened 2021-01-26
elementbase spec

brittany-johnson:
Handles parts of https://github.com/openstax/kitchen/issues/52

### DIFF
--- a/lib/kitchen/element_base.rb
+++ b/lib/kitchen/element_base.rb
@@ -162,7 +162,6 @@ module Kitchen
     #   element.set(:name,"div").set("id","foo")
     #
     def set(property, value)
-      # currently not being used outside tutorial
       case property.to_sym
       when :name
         self.name = value
@@ -182,7 +181,7 @@ module Kitchen
       @ancestors[type.to_sym]&.element || raise("No ancestor of type '#{type}'")
     end
 
-    # Returns true if this element has an ancestor of the given type
+    # Returns true iff this element has an ancestor of the given type
     #
     # @param type [String, Symbol] e.g. +:page+, +:term+
     # @return [Boolean]
--- a/lib/kitchen/element_base.rb
+++ b/lib/kitchen/element_base.rb
@@ -182,7 +182,7 @@ module Kitchen
       @ancestors[type.to_sym]&.element || raise("No ancestor of type '#{type}'")
     end
 
-    # Returns true iff this element has an ancestor of the given type
+    # Returns true if this element has an ancestor of the given type
     #
     # @param type [String, Symbol] e.g. +:page+, +:term+
     # @return [Boolean]
--- a/lib/kitchen/element_base.rb
+++ b/lib/kitchen/element_base.rb
@@ -162,6 +162,7 @@ module Kitchen
     #   element.set(:name,"div").set("id","foo")
     #
     def set(property, value)
+      # currently not being used outside tutorial
       case property.to_sym
       when :name
         self.name = value
--- a/spec/element_base_spec.rb
+++ b/spec/element_base_spec.rb
@@ -81,48 +81,14 @@ RSpec.describe Kitchen::ElementBase do
 
   describe '#has_ancestor?' do
     it 'returns true if element has ancestor of given type' do
+      p_element = element.chapters.pages.examples.search('p').first
+      expect(p_element.has_ancestor?(:chapter)).to eq true
     end
 
     it 'returns false if element does not have ancestor of given type' do
+      p_element = element.chapters.pages.examples.search('p').first
+      expect(p_element.has_ancestor?(:figure)).to eq false
     end
   end
-
-  context 'add_ancestors' do
-  end
-
-  describe '#add_ancestor' do
-  end
-
-  describe '#ancestor_elements' do
-  end
-
-  describe '#count_as_descendant' do
-  end
-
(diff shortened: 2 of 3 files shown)


spec/element_base_spec.rb
```diff
@@ -0,0 +1,94 @@
+require 'spec_helper'
+
+RSpec.describe Kitchen::ElementBase do
+
+  let(:element) do
```
jpslav, line 5: it is true that this returns an element (a `BookElement`).   Since everything is an element (including `para` below), might be more clear to call this `:book`

spec/element_base_spec.rb
```diff
@@ -0,0 +1,94 @@
...
+            <p>This is a paragraph.</p>
+          </div>
+        HTML
+      )
+    )
+  end
+
+  describe '#has_class?' do
+    it 'returns true if element has given class' do
+      div_element = element.search('div[data-type="example"]').first
```
jpslav, line 19: you can use the `first!` method instead: `element.first!('div[data-type="example"]')`  Ditto elsewhere.

`first!` blows up if the element isn't found, which is better here than returning `nil`.  The element should definitely be there, so better to know faster if you mistyped the search terms.

spec/element_base_spec.rb
```diff
@@ -0,0 +1,94 @@
...
+
+    it 'returns false if element does not have given class' do
+      div_element = element.search('div[data-type="example"]').first
+      expect(div_element.has_class?('class2')).to eq false
+    end
+  end
+
+  describe '#id' do
+    it 'returns the element\'s ID' do
+      div_element = element.search('div[data-type="example"]').first
```
jpslav, line 31: instead of repeating this line a few times, do this at the top after `let(:element)`

```ruby
let(:example) { element.first!('div[data-type="example"]') }
```

spec/element_base_spec.rb
```diff
@@ -0,0 +1,94 @@
...
+      expect(div_element.id).to eq 'div1'
+    end
+  end
+
+  describe '#id=' do
+    it 'sets the element\'s ID' do
+      para = element.search('p').first
```
jpslav, line 38:
```ruby
let(:para) { book.first!('p') }
```

and also use this below on line 56

spec/element_base_spec.rb
```diff
@@ -0,0 +1,94 @@
...
+      para.id = 'para1'
+      expect(para.id).to match('para1')
```
jpslav, line 40: Use `.to eq` instead of `.to match` because `match` will return true if there's a partial match, e.g.

```
irb(main):001:0> "sectionblah".match("section")
=> #<MatchData "section">
```

Ditto below.

spec/element_base_spec.rb
```diff
@@ -0,0 +1,94 @@
...
+        p_element.set(:name, 'span').set(:class, 'span1')
+      }.to change {
+        p_element.to_html
+      }.from(p_matcher).to(span_matcher)
+    end
+  end
+
+  describe '#ancestor' do
+    it 'returns element\'s ancestor of the given type' do
+      p_element = element.chapters.pages.examples.search('p').first
```
jpslav, line 69: `.first!('p')`

jpslav: Also define this with a `let` inside this `describe` block b/c it is the same in all three of these tests

```ruby
describe '#ancestor' do
  let(:p_element) { element.chapters.pages.examples.search('p').first }

  it '...' do
  ...
end
```

jpslav (changes requested):

lib/kitchen/element_base.rb
```diff
@@ -181,7 +182,7 @@ def ancestor(type)
       @ancestors[type.to_sym]&.element || raise("No ancestor of type '#{type}'")
     end
 
-    # Returns true iff this element has an ancestor of the given type
+    # Returns true if this element has an ancestor of the given type
```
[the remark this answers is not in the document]
brittany-johnson: oh oops! lol

spec/element_base_spec.rb
brittany-johnson: I change it to `book.chapters.pages.examples.search('p').first!` instead of `book.chapters.pages.examples.first!('p')` because when I tried to change it to the second, I got the error: 
```
ArgumentError:
       wrong number of arguments (given 1, expected 0)
```

I think this is because I'm chaining it to something from the ElementEnumerataor class and that class also has a `first!` method like ElementBase except it does not take an argument. 

This is just my guess/assumption, let me know if I'm completely off LOL! @jpslav

jpslav: ah ok you're right

spec/element_base_spec.rb
```diff
@@ -0,0 +1,90 @@
...
+
+  describe '#id' do
+    it 'returns the element\'s ID' do
+      expect(example.id).to eq 'div1'
+    end
+  end
+
+  describe '#id=' do
+    it 'sets the element\'s ID' do
+      para = book.first!('p')
```
jpslav, line 38: remove (you have this available via the `let` above)

jpslav (changes requested):

spec/element_base_spec.rb
brittany-johnson: ooooops

jpslav (approved):

PR #97, opened 2021-01-28
Create a spec for MoveTitleTextToSpan

### LINKED ISSUE #70: Write spec for MoveTitleTextIntoSpan

Stackblocks:
Resolves #70

### DIFF
--- a/spec/directions/move_title_text_into_span_spec.rb
+++ b/spec/directions/move_title_text_into_span_spec.rb
@@ -1,0 +1,24 @@
+require 'spec_helper'
+
+RSpec.describe Kitchen::Directions::MoveTitleTextIntoSpan do
+
+  let(:element_1) do
+    new_element(
+      <<~HTML
+        <h1 data-type="document-title">foo</h1>
+      HTML
+    )
+  end
+
+  it 'calls v1' do
+    title = element_1
+    described_class.v1(title: title)
+    expect(title).to match_normalized_html(
+      <<~HTML
+        <h1 data-type="document-title">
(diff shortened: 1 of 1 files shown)


spec/directions/move_title_text_into_span_spec.rb
```diff
@@ -0,0 +1,24 @@
+require 'spec_helper'
+
+RSpec.describe Kitchen::Directions::MoveTitleTextIntoSpan do
+
+  let(:element_1) do
```
jpslav, line 5: Make this be `let(:title)` and then you can get rid of `title = element_1` below

spec/directions/move_title_text_into_span_spec.rb
```diff
@@ -0,0 +1,24 @@
...
+    new_element(
+      <<~HTML
+        <h1 data-type="document-title">foo</h1>
+      HTML
+    )
+  end
+
+  it 'calls v1' do
```
jpslav, line 13: it is a little dumb but I often say `it 'works'` for a spec with one test that is testing everything.  Here you are testing more than it just calling v1, you are testing that when you call v1 it does what you want.

jpslav (changes requested):

spec/directions/move_title_text_into_span_spec.rb
Stackblocks: Glad to hear there's something as straightforward as `it 'works'`, I was honestly kind of stumped for what to put there 😆

jpslav (approved):

PR #103, opened 2021-02-01
Prepend the word 'chapter' to li elements in the toc answer key section

kswanson33:
Connected to #[3337](https://github.com/openstax/cnx-recipes/issues/3337)

### DIFF
--- a/lib/kitchen/directions/bake_toc.rb
+++ b/lib/kitchen/directions/bake_toc.rb
@@ -88,11 +88,6 @@ module Kitchen
           number.prepend(child: part_text.paste)
         end
 
-        # Targets the Answer Key
-        if page.has_ancestor?('composite_chapter') && page.ancestor('composite_chapter').search('span').first.text == 'Answer Key'
-          title.first('.os-text').prepend(child: 'Chapter ')
-        end
-
         <<~HTML
           <li class="#{li_class}" cnx-archive-shortid="" cnx-archive-uri="#{page.id}">
             <a href="##{page.id}">
--- a/lib/kitchen/directions/bake_toc.rb
+++ b/lib/kitchen/directions/bake_toc.rb
@@ -88,6 +88,11 @@ module Kitchen
           number.prepend(child: part_text.paste)
         end
 
+        # Targets the Answer Key
+        if page.has_ancestor?('composite_chapter') && page.ancestor('composite_chapter').search('span').first.text == 'Answer Key'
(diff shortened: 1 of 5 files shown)


lib/kitchen/directions/bake_toc.rb
```diff
@@ -88,6 +88,11 @@ def self.li_for_page(page)
           number.prepend(child: part_text.paste)
         end
 
+        # Targets the Answer Key
+        if page.has_ancestor?('composite_chapter') && page.ancestor('composite_chapter').search('span').first.text == 'Answer Key'
```
kswanson33, line 92: Is there a better way to get at the correct pages?

lib/kitchen/directions/bake_toc.rb
```diff
@@ -88,6 +88,11 @@ def self.li_for_page(page)
...
+          title.first('.os-text').prepend(child: 'Chapter ')
```
kswanson33, line 93: Currently, 'Chapter ' is a hard-coded string; is there a better way to do this?

jpslav: Use I18n translations.  See [`BakeExample`](https://github.com/openstax/kitchen/blob/master/lib/kitchen/directions/bake_example.rb#L16) which pulls in the translation for the `example_label` key in [`lib/locales/en.yml`](https://github.com/openstax/kitchen/blob/master/lib/locales/en.yml#L5).

jpslav: I'm wondering if Chapter is added in the baked version you have, I wonder if it is also in the answer key page title and can just be pulled in that way.

lib/kitchen/directions/bake_toc.rb
jpslav: let's look at this together

lib/kitchen/directions/bake_toc.rb
kswanson33: Yes, it looks like Chapter is added in the answer key page, and that it would work to just pull that into the ToC. I don't see a `BakeAnswerKey` class, so I'm going to go ahead and try to get one working, and you can tell me when you have a chance if I'm barking up the wrong tree 😅

kswanson33:
This now fixes the missing word "chapter" by adding it into the Answer Key. One thing to note is that it's now important that the TOC be baked after the Answer Key in order for the word to be pulled into the TOC titles. How much should I worry about that?

jpslav (approved):

Issue #44, opened 2021-01-14
Create and test `BakePreface` direction


jpslav:
Move [code from the Chemistry 2e recipe](https://github.com/openstax/recipes/blob/main/books/chemistry2e/bake#L19-L27) that bakes preface titles into a new `BakePreface` direction.  Add tests.

Do this in the style of [`BakeFootnotes`](https://github.com/openstax/kitchen/tree/master/lib/kitchen/directions/bake_footnotes), i.e.

1. Make a `lib/kitchen/directions/bake_preface` directory
2. Add `main.rb` and `v1.rb` files like in `bake_footnotes`
3. Add `spec/directions/bake_preface/v1_spec.rb` to contain the test, see the footnotes version of this as a guide.

Note that [this line](https://github.com/openstax/recipes/blob/main/books/chemistry2e/bake#L20) iterates over titles in the preface.  I think this is because there is also a title element in the hidden metadata -- legacy baking changed both, so so did I.  At some future point maybe we won't care to change both, but for now we're trying to do everything like legacy did.

So add a `titles` method to `PageElement`.  It'll look a lot like the existing [singular `title` method](https://github.com/openstax/kitchen/blob/master/lib/kitchen/page_element.rb#L24-L27) but will not return just one title.

PR #104, opened 2021-02-01
Bake preface

brittany-johnson:
PR 1 of 2 for #44 

Next, I need to remove [this block](https://github.com/openstax/recipes/blob/main/books/chemistry2e/bake#L19-L27) from the chem recipe and add the BakePreface

### DIFF
--- a/lib/kitchen/page_element.rb
+++ b/lib/kitchen/page_element.rb
@@ -26,6 +26,14 @@ module Kitchen
       first!(is_introduction? ? selectors.title_in_introduction_page : selectors.title_in_page)
     end
 
+    # Returns an enumerator for titles.
+    #
+    # @return [ElementEnumerator]
+    #
+    def titles
+      search("div[data-type='document-title']")
+    end
+
     # Returns true if this page is an introduction
     #
     # @return [Boolean]
--- a/spec/page_element_spec.rb
+++ b/spec/page_element_spec.rb
@@ -22,7 +22,7 @@ RSpec.describe Kitchen::PageElement do
       )).pages.first
   end
 
-  let(:page2) do
+  let(:page_with_multiple_titles) do
     book_containing(html:
       <<~HTML
         <div data-type="page">
@@ -32,7 +32,7 @@ RSpec.describe Kitchen::PageElement do
           </div>
         </div>
       HTML
-    )
+    ).pages.first!
   end
 
   describe '#title' do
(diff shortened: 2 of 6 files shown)


lib/kitchen/page_element.rb
```diff
@@ -26,6 +26,14 @@ def title
       first!(is_introduction? ? selectors.title_in_introduction_page : selectors.title_in_page)
     end
 
+    # Returns an enumerator for titles.
+    #
+    # @return [ElementEnumerator]
+    #
+    def titles
+      search("div[data-type='document-title']")
+    end
```
jpslav, line 35: Please add a spec for this in page_element_spec.rb

jpslav (changes requested):

spec/page_element_spec.rb
```diff
@@ -22,20 +22,41 @@
...
+    book_containing(html:
+      <<~HTML
+        <div data-type="page">
+          <div data-type="document-title">Title</div>
+          <div data-type="metadata">
+            <div data-type="document-title">Title MetaData</div>
+          </div>
+        </div>
+      HTML
+    )
```
jpslav, line 35: I would not name this variable `page2` because it is actually a book.  Leads to `page2.pages[0]` below which looks funny.  

Instead what I would do is name it some more meaningful and drill down to the page here, e.g.

```
let(:page_with_multiple_titles) do 
  book_containing(html:
    <<~HTML
      <div data-type="page">
        <div data-type="document-title">Title</div>
        <div data-type="metadata">
          <div data-type="document-title">Title MetaData</div>
        </div>
      </div>
    HTML
  ).pages.first!
end
```

spec/page_element_spec.rb
```diff
@@ -22,20 +22,41 @@
...
+        expect(page1.title.text).to eq page_title_text
       end
     end
   end
 
+  describe '#titles' do
+    it 'returns all titles in element' do
+      page_titles = '<div data-type="document-title">Title</div>'\
+      '<div data-type="document-title">Title MetaData</div>'
+      expect(page2.pages[0].titles.to_s).to eq page_titles
```
jpslav, line 58: See comment about above a different variable name and then you can simplify these three lines to

```ruby
expect(page_with_multiple_titles.titles.to_s).to match(/Title.*Title Metadata/)
```

or 

```ruby
expect(page_with_multiple_titles.titles.map(&:to_s)).to match([/Title/, /Title MetaData/])
```

using a regular expression(s) to match the parts you care about

jpslav (changes requested):

jpslav (approved):

PR #105, opened 2021-02-01
Add class to outer wrapper for tables

kswanson33:
Related to [3337](https://github.com/openstax/cnx-recipes/issues/3337)

### DIFF
--- a/lib/kitchen/directions/bake_numbered_table/v1.rb
+++ b/lib/kitchen/directions/bake_numbered_table/v1.rb
@@ -2,19 +2,19 @@ module Kitchen::Directions::BakeNumberedTable
   class V1
 
     def bake(table:, number:)
+      table.wrap(%(<div class="os-table">))
+
       table_label = "#{I18n.t(:table_label)} #{number}"
       table.document.pantry(name: :link_text).store table_label, label: table.id
 
       if table.top_titled?
-        table.wrap(%(<div class="os-table os-top-titled-container">))
+        table.parent.add_class('os-top-titled-container')
         table.prepend(sibling:
           <<~HTML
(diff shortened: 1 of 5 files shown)


lib/kitchen/directions/bake_numbered_table/v1.rb
```diff
@@ -2,18 +2,19 @@ module Kitchen::Directions::BakeNumberedTable
   class V1
 
     def bake(table:, number:)
-      table.wrap(%(<div class="os-table">))
-
       table_label = "#{I18n.t(:table_label)} #{number}"
       table.document.pantry(name: :link_text).store table_label, label: table.id
 
       if table.top_titled?
+        table.wrap(%(<div class="os-table os-top-titled-container">))
```
jpslav, line 9: instead of having the `.wrap` call in two places, you can leave it where it is above and then do `table.parent.add_class("os-top-titled-container")`

jpslav (changes requested):

lib/kitchen/directions/bake_unnumbered_tables.rb
```diff
@@ -5,6 +5,8 @@ def self.v1(book:)
         book.tables('$.unnumbered').each do |table|
           table.wrap(%(<div class="os-table">))
           table.remove_attribute('summary')
+          table.unstyled? ? table.parent.add_class('os-unstyled-container') : 0
```
jpslav, line 8: like the one-liner but you can get by without the false `0` unused return value with

```ruby
table.parent.add_class('os-unstyled-container') if table.unstyled?
```

jpslav (changes requested):

jpslav (approved):

PR #93, opened 2021-01-27
Adds spec for bake chapter key equations

KarinaMendez2:
for: https://app.zenhub.com/workspaces/ce-styles-5cdc84baee941c07853fbbf2/issues/openstax/kitchen/59

PR #106, opened 2021-02-01
Bake chapter summary spec

LostMojo:


spec/directions/bake_chapter_summary_spec.rb
```diff
@@ -0,0 +1,82 @@
...
+              <h3 data-type='title'>Summary</h3>
+              <p>Ooh, it's another bit of text.</p>
+          </section>
+        </div>
+      HTML
+    )
+  end
+
+  #TODO: check that it:
+  #Replaces title with h3 title (children: os-number, os-divider, os-text)
```
jpslav, line 27: Is this TODO TODONE?

spec/directions/bake_chapter_summary_spec.rb
```diff
@@ -0,0 +1,82 @@
+require 'spec_helper'
+
+RSpec.describe Kitchen::Directions::BakeChapterSummary do
+
+  let(:chapter) do
+    chapter_element(
+      <<~HTML
+        <div data-type='page' id ="01">
```
jpslav, line 8: Add an introduction page to test that `chapter.pages('$:not(.introduction)')` is not iterating over it.

spec/directions/bake_chapter_summary_spec.rb
```diff
@@ -0,0 +1,82 @@
...
+    )
+  end
+
+  #TODO: check that it:
+  #Replaces title with h3 title (children: os-number, os-divider, os-text)
+  #
+
+  context 'when v1 is called on a chapter' do
+    it 'works' do
+      described_class.v1(chapter: chapter, metadata_source: metadata_element)
```
jpslav, line 32: Modify `metadata_element` to include an extra child that does not get copied along with the others, to test that the `metadata_source.search(...)` line only gets the parts we want.  I asked Karina to do this in another PR, see https://github.com/openstax/kitchen/pull/93/files#diff-d3c2c4afb9a47770b8067894bffac4b7e53477ae34b3e37782644c522616c04cR30

spec/directions/bake_chapter_summary_spec.rb
```diff
@@ -0,0 +1,82 @@
+require 'spec_helper'
+
+RSpec.describe Kitchen::Directions::BakeChapterSummary do
+
+  let(:chapter) do
+    chapter_element(
+      <<~HTML
+        <div data-type='page' id ="01">
+          <h1 data-type="document-title" itemprop="name">First Title</h1>
```
jpslav, line 9: Add an ID on this title to see that it gets populated in the `<a href="#that_id">` tag.  Ditto for the second page.

spec/directions/bake_chapter_summary_spec.rb
```diff
@@ -0,0 +1,82 @@
...
+          <div data-type="chapter">
+            <div data-type="page" id="01">
+              <h1 data-type="document-title" itemprop="name">First Title</h1>
+            </div>
+            <div data-type="page" id="02">
+              <h1 data-type="document-title" itemprop="name">Second Title</h1>
+            </div>
+            <div class="os-eoc os-summary-container" data-type="composite-page" data-uuid-key=".summary">
+              <h2 data-type="document-title">
+                <span class="os-text">Summary</span>
```
jpslav, line 46: This "Summary" text is pulled from `lib/locales/en.yml`, a set of English translations of certain words.  For our Spanish books, we'll have an `es.yml` or other file that has "Summary" translated into Spanish.  All words that we insert into books should be in these "locales" files for internationalization (aka "i18n" == "i followed by 18 letters followed by n" == "internationalization").

One day we may decide to change the English word we use here but we don't want this spec to break because of it.  To protect ourselves against that, we tell rspec to use a fake translation for this term with this block at the top of this file:

```ruby
before do
  stub_locales({
    'eoc_summary_title': 'Summary'
  })
end
```

You can see this in action [here in another spec](https://github.com/openstax/kitchen/blob/master/spec/directions/bake_numbered_table/v1_spec.rb#L5-L9).

Without this, if you change the value for the `eoc_summary_title` in `en.yml`, this test would break.  With this, such changes do not break the spec.

Read a tiny bit more about i18n [in the README](https://github.com/openstax/kitchen#internationalization-i18n).

jpslav (changes requested):

jpslav (approved):

jpslav:
Thanks @LostMojo!

PR #114, opened 2021-02-05
Add a Learning Objectives h3 header to content modules

kswanson33:
Related to openstax/cnx-recipes#3337

### DIFF
--- a/lib/kitchen/chapter_element.rb
+++ b/lib/kitchen/chapter_element.rb
@@ -1,3 +1,5 @@
+# frozen_string_literal: true
+
 module Kitchen
   # An element for a chapter
   #
@@ -22,7 +24,8 @@ module Kitchen
     #
     def title
       # Get the title in the immediate children, not the one in the metadata.  Could use
-      # CSS of ":not([data-type='metadata']) > [data-type='document-title'], [data-type='document-title']"
+      # CSS of ":not([data-type='metadata']) >
+      #       [data-type='document-title'], [data-type='document-title']"
       # but xpath is shorter
       first!("./*[@data-type = 'document-title']")
     end
@@ -49,6 +52,14 @@ module Kitchen
     #
     def key_equations
       search('section.key-equations')
+    end
+
+    # Returns an enumerator for the abstracts
+    #
+    # @return [ElementEnumerator]
+    #
+    def abstracts
+      search('[data-type="abstract"]')
     end
 
     # Returns true if this class represents the element for the given node
--- a/lib/kitchen/directions/bake_page_abstracts.rb
+++ b/lib/kitchen/directions/bake_page_abstracts.rb
@@ -1,0 +1,14 @@
+# frozen_string_literal: true
+
+module Kitchen
+  module Directions
+    # Adds learning objectives header to abstracts
+    module BakePageAbstracts
+      def self.v1(chapter:)
+        chapter.abstracts.each do |abstract|
+          abstract.prepend(child: "<h3 data-type='title'>#{I18n.t(:learning_objectives)}</h3>")
+        end
+      end
+    end
+  end
+end
--- a/lib/kitchen/page_element.rb
+++ b/lib/kitchen/page_element.rb
@@ -1,3 +1,5 @@
+# frozen_string_literal: true
+
 module Kitchen
   # An element for a page
   #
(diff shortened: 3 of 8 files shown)


kswanson33:
@jpslav I've implemented this as a separate class, but the `BakeChapterIntroductions` class also iterates through the chapter content modules. Would it be better to make this a part of `BakeChapterIntroductions`, or is it possible not all books will have the Learning Objectives header?

jpslav:
@kswanson33 When I look at the original baked output, I see that introduction pages no longer have a `data-type="abstract"` element, which matches [this line in BakeChapterIntroductions](https://github.com/openstax/kitchen/blob/master/lib/kitchen/directions/bake_chapter_introductions.rb#L23).  So I think you are right to put this code in a separate class and not in BakeChapterIntroductions.  In general I'm a fan of calling more smaller directions than fewer larger ones, b/c as you note, maybe there are some other books that don't have this learning objectives header.

lib/kitchen/directions/bake_chapter_content_module.rb
```diff
@@ -0,0 +1,13 @@
+module Kitchen
+  module Directions
+    module BakeChapterContentModule
```
jpslav, line 3: The thing that is being baked here is the abstract in a page.  So I think the name of this class should reflect that somewhat.  Even just `BakeAbstracts` or `BakePageAbstracts`.  Note also that the recipes already iterate through chapters, so if you wanted to leave this as working on a chapter, your method could be `def self.v1(chapter:)` and then call this direction from inside a chapter loop in the recipe.  This latter part is your call, whatever makes the most sense.  

Are there ever abstracts that are not in chapters?  If so `book.search('[data-type="abstract"]')` would be better.  There may not be tho.

lib/kitchen/directions/bake_chapter_content_module.rb
```diff
@@ -0,0 +1,13 @@
...
+      def self.v1(book:)
+        book.chapters.each do |chapter|
+          chapter.search('[data-type="abstract"]').each do |abstract|
```
jpslav, line 6: wherever you end up putting the search, on chapter, page, or book, make an `abstract` or `abstracts` method to encapsulate this CSS selector, e.g. https://github.com/openstax/kitchen/blob/master/lib/kitchen/chapter_element.rb#L42-L44.  Also make a spec for that method.

lib/kitchen/directions/bake_chapter_content_module.rb
```diff
@@ -0,0 +1,13 @@
...
+            abstract.prepend(child:'<h3 data-type="title">Learning Objectives</h3>')
```
jpslav, line 7: Use an I18n translation here instead of hardcoding the text.  Make sure to stub the translation in the spec

lib/kitchen/directions/bake_chapter_content_module.rb
```diff
@@ -0,0 +1,13 @@
+module Kitchen
```
jpslav, line 1: are rubocop and lefthook running?  thought it would have made you put in module comments and a frozen string literal magic comment.

lib/kitchen/directions/bake_chapter_content_module.rb
jpslav: This may be moot depending on how you implement other parts of this review, but in this case where `chapter` is only used once, you don't need to get each chapter and then search on it -- you can just search on `.chapters`, like

```ruby
book.chapters.search(...).each do |abstract|
```

jpslav (changes requested):

lib/kitchen/directions/bake_chapter_content_module.rb
kswanson33: There are not abstracts (at least, not ones that require this insertion) outside of chapters, at least not in chemistry.

kswanson33:
@jpslav I added that method you mentioned to the `chapter` element, but didn't add a spec -- there isn't, in fact, a spec for the chapter element on its own. Codecov doesn't seem to register the lack of coverage, either. Any ideas about why this is happening?

jpslav (approved):

jpslav:
The new method is used in the spec of the direction (where it says `chapter.abstracts`).  So it has coverage

PR #118, opened 2021-02-05
Use MathML namespace in specs

kswanson33:


spec/directions/bake_exercises_spec.rb
```diff
@@ -77,131 +77,129 @@
   it 'works' do
     described_class.v1(book: book_1)
 
-    expect(book_1.to_s).to match_normalized_html(
+    expect(book_1.body.to_s).to match_normalized_html(
```
jpslav, line 80: minor: remove `to_s`

jpslav, line 80: minor: remove .to_s

lib/kitchen/directions/bake_math_in_paragraph.rb
```diff
@@ -2,7 +2,7 @@ module Kitchen
   module Directions
     module BakeMathInParagraph
       def self.v1(book:)
-        book.search('//p//math | //p//m').each do |math|
+        book.search('p m|math', '//p//math | //p//m').each do |math|
```
jpslav, line 5: rubocop / lefthook are maybe not running because this shouldn't have been pushable without the frozen string literal comment at the top.

jpslav (changes requested):

lib/kitchen/directions/bake_math_in_paragraph.rb
kswanson33: Rubocop is running, but it hasn't been forcing me to resolve all issues before pushing

jpslav (approved):

PR #124, opened 2021-02-08
Add specs for document 

KarinaMendez2:
for: https://app.zenhub.com/workspaces/ce-styles-5cdc84baee941c07853fbbf2/issues/openstax/kitchen/123

### DIFF
--- a/lib/kitchen/document.rb
+++ b/lib/kitchen/document.rb
@@ -20,7 +20,6 @@ module Kitchen
     #   @return [Selectors::Base]
     def_delegators :config, :selectors
 
-    # rubocop:disable Layout/LineLength
     # @!method to_xhtml
     #   @see https://www.rubydoc.info/github/sparklemotion/nokogiri/Nokogiri/XML/Node#to_xhtml-instance_method Nokogiri::XML::Node#to_xhtml
     #   @return [String] the document as an XHTML string
@@ -36,7 +35,6 @@ module Kitchen
     #   @see https://www.rubydoc.info/github/sparklemotion/nokogiri/Nokogiri/XML/Node#to_html-instance_method Nokogiri::XML::Node#to_html
     #   @return [String] the document as an HTML string
     def_delegators :@nokogiri_document, :to_xhtml, :to_s, :to_xml, :to_html
-    # rubocop:enable Layout/LineLength
 
     # Return a new instance of Document
     #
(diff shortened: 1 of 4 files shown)


lib/kitchen/document.rb
```diff
@@ -18,6 +20,7 @@ class Document
     #   @return [Selectors::Base]
     def_delegators :config, :selectors
 
+    # rubocop:disable Layout/LineLength
```
jpslav, line 23: You can add a regular expression to ignore this cop for lines with comments (in .rubocop.yml)

https://stackoverflow.com/questions/26342570/rubocop-linelength-how-to-ignore-lines-with-comments#comment81397139_42833040

```yaml
Layout/LineLength:
  Max: 80
  IgnoredPatterns: ['(\A|\s)#']
```

@kswanson33 just had this problem in [her PR](https://github.com/openstax/kitchen/pull/114/files#diff-65a4951dc80fa8de7805fc7491d956d6982228167804db2e3ff0ae87c011d204R27) where she had to do a funny line break for a comment with code.  So I think it'd be good to turn it off for comments.

Then you can get rid of these `disable` / `enable` directives.

lib/kitchen/document.rb
```diff
@@ -131,11 +137,7 @@ def create_element(name, *args, &block)
     # @return [Element]
     #
     def create_element_from_string(string)
-      Kitchen::Element.new(
-        node: @nokogiri_document.create_element(string),
-        document: self,
-        short_type: "created_element_#{SecureRandom.hex(4)}"
-      ).element_children.first
+      new_element(html: string)
```
jpslav, line 140: Is this calling the spec helper?  Don't do that.  Specs can depend on real code but real code should not depend on specs.

spec/document_spec.rb
```diff
@@ -0,0 +1,39 @@
...
+    new_element(
+      <<~HTML
+        <div>
+          <div data-type="chapter" id="chapter1">
+            <span>This is a chapter</span>
+          </div>
+        </div>
+      HTML
+    )
+  end
```
jpslav, line 16: `book1` is actually a `BookDocument`, which yes is a `Document` but I think it'd be better to just instantiate a `Kitchen::Document` directly with a dummy Nokogiri document.  

Also below is saying `book1.document` everywhere -- `book1` is never used so it shouldn't be what is returned, better to have `let(:document)`

jpslav (changes requested):

jpslav (approved):

PR #125, opened 2021-02-08
spec for #add_ancestor

brittany-johnson:
part of #52

### DIFF
--- a/spec/element_base_spec.rb
+++ b/spec/element_base_spec.rb
@@ -110,6 +110,13 @@ RSpec.describe Kitchen::ElementBase do
       expect(para).to receive(:add_ancestor).with(ancestor_example)
       para.add_ancestors(ancestor_example)
     end
+
+    it 'raises an error if unsupported ancestor type given' do
+      random_string = 'blah'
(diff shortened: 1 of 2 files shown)


spec/element_base_spec.rb
```diff
@@ -87,6 +87,24 @@
     end
   end
 
+  describe '#add_ancestor' do
```
brittany-johnson, line 90: I only did `#add_ancestor` and not `#add_ancestors` because I thought that since in all cases, `#add_ancestors` uses `#add_ancestor`, a separate spec for `#add_ancestors` isn't necessary. Is this correct? @jpslav

jpslav (approved):
@brittany-johnson I'm approving this one so that we can merge the weird rspec failure fix, but we'll need to come back and add coverage for [the Hash case](https://github.com/openstax/kitchen/blob/master/lib/kitchen/element_base.rb#L208-L209) and [the Element/Document case](https://github.com/openstax/kitchen/blob/master/lib/kitchen/element_base.rb#L212-L213)

PR #133, opened 2021-02-12
Refactor metadata into new class MetadataElement

kswanson33:
related to openstax/cnx-recipes#3337

### DIFF
--- a/lib/kitchen/directions/bake_index/v1.rb
+++ b/lib/kitchen/directions/bake_index/v1.rb
@@ -1,3 +1,5 @@
+# frozen_string_literal: true
+
 module Kitchen::Directions::BakeIndex
   class V1
     renderable
@@ -111,8 +113,7 @@ module Kitchen::Directions::BakeIndex
     end
 
     def bake(book:)
-      @metadata_elements = book.metadata.search(%w([data-type='revised'] .authors .publishers .print-style
-                                                   .permissions [data-type='subject'])).copy
+      @metadata_elements = book.metadata.data.copy
       @index = Index.new
 
       book.pages.terms.each do |term_element|
--- a/lib/kitchen/directions/bake_index/v1.rb
+++ b/lib/kitchen/directions/bake_index/v1.rb
@@ -1,149 +1,145 @@
 # frozen_string_literal: true
 
-module Kitchen
-  module Directions
-    module BakeIndex
-      # Bake directions for eob index
-      #
-      class V1
-        renderable
+module Kitchen::Directions::BakeIndex
+  # Bake directions for eob index
+  #
+  class V1
+    renderable
 
-        class Term
-          attr_reader :text
-          attr_reader :id
-          attr_reader :group_by
-          attr_reader :page_title
+    class Term
+      attr_reader :text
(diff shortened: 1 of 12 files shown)


lib/kitchen/directions/bake_index/v1.rb
```diff
@@ -1,142 +1,149 @@
...
-        @id = id
-        @group_by = group_by
-        @page_title = page_title
-      end
-    end
+# frozen_string_literal: true
+
+module Kitchen
+  module Directions
+    module BakeIndex
```
jpslav, line 5: I know Rubocop says to do it this way instead of `Kitchen::Directions::BakeIndex` but I'm not a huge fan b/c it loses like 10 characters for so much indenting.  What about ignoring this rule at least in directions which have a lot of nesting?

lib/kitchen/element_base.rb
```diff
@@ -590,7 +592,8 @@ def clone
 
     # @!method pages
     #   Returns a pages enumerator
-    def_delegators :as_enumerator, :pages, :chapters, :terms, :figures, :notes, :tables, :examples
+    def_delegators :as_enumerator, :pages, :chapters, :terms, :figures, :notes, :tables, :examples,
+                   :metadatas # doing my best with the plural of data :/
```
jpslav, line 596: that works, you can remove the comment

lib/kitchen/metadata_element.rb
```diff
@@ -0,0 +1,28 @@
...
+            document: document,
+            enumerator_class: MetadataElementEnumerator,
+            short_type: :metadata)
+    end
+
+    # Returns set of selected data elements
+    #
+    # @return [ElementEnumerator]
+    #
+    def data
```
jpslav, line 23: Is there a more meaningful name we can use?  Like `elements_to_keep`? (that name isn't awesome either but it conveys that these are child elements that we want to persist through baking.

spec/helpers/factory_helpers.rb
```diff
@@ -65,5 +65,5 @@ def metadata_element
         <div data-type="subject" id="subject">Subject</div>
       </div>
     HTML
-  )
+  ).metadata
```
jpslav, line 68: why this change?

jpslav (changes requested):

spec/helpers/factory_helpers.rb
kswanson33: Specs fail without this change -- several functions (i.e. `BakeChapterGlossary.v1`)  have been updated to expect a `MetadataElement` instead of an `Element`. This function is called in the specs to generate metadata, so it must now also return a `MetadataElement`.

lib/kitchen/metadata_element.rb
kswanson33: What about these?: `data_to_keep` / `children_to_keep` / `descendants_to_keep`

spec/helpers/factory_helpers.rb
jpslav: ah right

jpslav (approved):

PR #134, opened 2021-02-12
Add BakeLinkPlaceholders class to base kitchen

kswanson33:
Related to openstax/cnx-recipes#3337

Refactors https://github.com/openstax/recipes/blob/main/books/chemistry/bake#L113:L125 and moves highlighted section into Kitchen.

### DIFF
--- a/lib/kitchen/directions/bake_link_placeholders.rb
+++ b/lib/kitchen/directions/bake_link_placeholders.rb
@@ -1,0 +1,24 @@
+# frozen_string_literal: true
+
+module Kitchen
+  module Directions
+    # Bake directions for link placeholders
+    #
+    module BakeLinkPlaceholders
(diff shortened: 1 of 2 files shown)


lib/kitchen/directions/bake_link_placeholders.rb
```diff
@@ -0,0 +1,24 @@
+# frozen_string_literal: true
+
+module Kitchen
+  module Directions
+    # Bake directions for link placeholders
+    #
+    module BakeLinkPlaceholders
+      def self.v1(book:, pantry:)
```
jpslav, line 8: `pantry` should be the same as `book.document.pantry` so can probably nix the second argument

jpslav (changes requested):

jpslav (approved):

PR #137, opened 2021-02-22
Unit TOC

brittany-johnson:
Need for https://github.com/openstax/cnx-recipes/issues/3506 and all other books w/ units


lib/kitchen/directions/bake_chapter_summary.rb
```diff
@@ -10,7 +10,8 @@ def self.v1(chapter:, metadata_source:)
         # TODO: include specific page types somehow without writing it out
         chapter.pages('$:not(.introduction)').each do |page|
           summary = page.summary
-          summary.first("[data-type='title']").trash # get rid of old title
+          summary.first("[data-type='title']").trash if
+            summary.first("[data-type='title']").respond_to? :trash # get rid of old title if exists
```
brittany-johnson, line 14: in am gov 2e, not all summaries have a title

This is not related to unit toc so if you prefer I can move out to a different PR 

Also wondering if this is the clearest way to write this?

brittany-johnson:
The codecov report isn't displaying properly on this PR but I saw locally that this is the line that I'm missing. I'm not sure how to test this though: 
Screen Shot 2021-03-10 at 4 36 46 PM

lib/kitchen/directions/bake_chapter_summary.rb
jpslav:
```ruby
summary.first("[data-type='title']")&.trash
```

The `&` is the [safe navigation operator](https://mitrev.net/ruby/2015/11/13/the-operator-in-ruby/) in Ruby.  It'll only call the next method if the part to the left of it is not `nil`.

Fine to have a random change here I think.

lib/kitchen/unit_element.rb
```diff
@@ -0,0 +1,37 @@
...
+    # @param document [Document] this element's document
+    #
+    def initialize(node:, document: nil)
+      super(node: node,
+            document: document,
+            enumerator_class: UnitElementEnumerator,
+            short_type: :unit)
+    end
+
+    def title
```
jpslav, line 20: Add method documentation here (the YARD docs like above and below)

spec/directions/bake_toc_spec.rb
```diff
@@ -19,51 +20,86 @@
             <span data-type="" itemprop="" class="os-text">Preface</span>
           </h1>
         </div>
-        <div data-type="chapter">
```
jpslav, line 22: Can you leave the unit-less `book1` and its spec in and then add another spec that shows that units also work?  That way we can test both; as is the test no longer tests the case without units.

jpslav (changes requested):

spec/directions/bake_toc_spec.rb
brittany-johnson: oop and this fixed the code coverage issue I was having 😆

lib/kitchen/unit_element.rb
jpslav: Needs 

```
# @return [Element]
```

brittany-johnson: 🙈

jpslav (approved):

PR #138, opened 2021-02-25
Bake Exercises methods

KarinaMendez2:
for: https://app.zenhub.com/workspaces/ce-styles-5cdc84baee941c07853fbbf2/issues/openstax/cnx-recipes/3376


kswanson33 (changes requested):
I think this is starting to work for a lot of different types of books, which is good! But I'm thinking about how I'd call this from a recipe, and seeing some room for refactoring.

Right now, in order to use these functions for a book with 2 exercises sections, I would bake it like this:

```ruby
BakeExercises.v1(book: book, class_name: '<name1>', bake_eob: false)
BakeExercises.v1(book: book, class_name: '<name2>', bake_eob: false)
BakeExercises.eob(book: book, class_names:
    {
      '<title_key_1>':
        {
          classname: '<name1>',
          decimal: false,
          sectional: false
        },
      '<title_key_2>':
        {
          classname: '<name2>',
          decimal: false,
          sectional: false
        },
    })
```

There's a lot for the developer to remember here, & it doesn't feel very modular. Instead, I can see a few possible alternatives.

**Option 1:**

```ruby
BakeExercises.problems(book: book, selectors: [name1, name2])
BakeExercises.solutions(book: book, selectors: {title_key_1: name1, title_key_2: name2})
# extra logic for numbering exercises if needed
```

Depending on whether `selectors` is given multiple values or a single value, the logic for baking one or multiple exercise/solutions sections can be inferred.

**Option 2:**

```ruby
BakeExercises.problems(book: book, selector: name1)
BakeExercises.problems(book: book, selector: name2)
BakeExercises.solutions(book: book, selector: name1, title: title_key_1)
BakeExercises.solutions(book: book, selector: name2, title: title_key_2)
# extra logic for numbering exercises if needed
```

This would assume that `BakeExercises.problems` always expects to bake a single exercise section, as does `BakeExercises.solutions`. If a book has more than one section, these methods are called multiple times.

**Option 3:**

```ruby
# for a book like chemistry
BakeExercises.bake_single_exercise_section(book: book, selector: name1, title: title_key_1)
```

```ruby
# for calculus, university physics, etc
BakeExercises.bake_multiple_exercise_sections(book: book, selector: [name1, name2, ...], title: [title_key_1, title_key_2, ...])
```

This would have two functions (sort of like v1/v2/etc), that would treat books with single exercise sections vs books with multiple exercise sections as two different cases.

Does that make sense? These options aren't the only possibilities. I'm mainly trying to illustrate that, if possible, I'd love 1) for these functions to be called with a more parallel structure, 2) to avoid passing in booleans as arguments to control the flow of a function, and 3) to either completely separate or completely combine the logic for baking problems and solutions (instead of the current way, which is to have half the solutions logic in V1 and the other half in EOB). 

It looks really promising overall. I think we're going to be able to figure out how to make this work for most books. Thanks for digging into this!

lib/locales/en.yml
```diff
@@ -3,21 +3,27 @@ en:
...
   exercise_label: Exercise
   chapter: Chapter
+  checkpoint: Checkpoint
   unit: Unit
   learning_objectives: Learning Objectives
   stepwise_step_label: Step
   eoc_key_terms_title: Key Terms
   eoc_summary_title: Summary
-  eoc_exercises_title: Exercises
+  eoc_exercises_title: Review Exercises
```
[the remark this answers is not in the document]
jpslav: We need book-specific translations I think, likely that live in recipes

jpslav (changes requested):
@KarinaMendez2 great job on a difficult task! As you'll see I have recommended moving some of the code into its own direction class.  I also made another PR into this PR with implementations to do that (helped me think through it to code it, this isn't one of those PRs where you can convey everything in comments).  I also have a review going of the recipe and will submit that, and I may also make a sketch recipe PR that gives approaches for how to solve the bigger numbering challenges -- that'll make use of some of the reorganizations recommended here.

lib/kitchen/directions/bake_notes.rb
```diff
@@ -46,7 +46,7 @@ def self.v1(book:)
             note.prepend(child:
               <<~HTML
                 <h3 class="os-title" data-type="title">
-                  <span class="os-title-label">#{note.autogenerated_title}</span>
+                  <span class="os-title-label">#{note.has_class?('media-2') ? 'Media' : note.autogenerated_title}</span>
```
kswanson33, line 49: just FYI university physics has a note with the `media-2` class, except the title is `Interactive` 👀

lib/kitchen/directions/bake_chapter_answer_key/strategies/calculus.rb
```diff
@@ -0,0 +1,38 @@
+# frozen_string_literal: true
+
+module Kitchen::Directions::BakeChapterAnswerKey
```
kswanson33, line 3: Since the end of chapter doesn't contain an answer key, could this module be renamed to something along the lines of `BakeChapterExercises`?

lib/kitchen/directions/bake_notes.rb
```diff
@@ -4,7 +4,7 @@ module Kitchen
   module Directions
     module BakeNotes
       def self.v1(book:)
-        book.notes.each do |note|
+        book.notes('$:not(.checkpoint):not(.theorem)').each do |note|
```
kswanson33, line 7: The filtered classes `checkpoint` and `theorem` may be calculus specific. I think we're going to need a larger-scale notes refactor -- I'll probably do that in another PR shortly.

lib/kitchen/directions/bake_theorem/v1.rb
```diff
@@ -0,0 +1,28 @@
+# frozen_string_literal: true
+
+module Kitchen::Directions::BakeTheorem
+  class V1
+    def bake(theorem:, number:)
+      theorem['use-subtitle'] = true
+      new_subtitle = theorem.title.cut
+
+      theorem.replace_children(with: %(<div class="os-note-body">#{theorem.children}</div>))
+      note_body = theorem.first('.os-note-body')
```
kswanson33, line 10: Is `theorem` a type of note? This direction should probably be a variety of `BakeNotes` -- see my comment about notes refactoring

lib/kitchen/directions/bake_book_answer_key/eob.xhtml.erb
```diff
@@ -0,0 +1,9 @@
+<div class="os-eob os-solutions-container" data-type="composite-chapter" data-uuid-key=".solutions">
+  <h1 data-type="document-title" id="composite-chapter-1">
+    <span class="os-text"><%= I18n.t(:eoc_answer_key_title) %></span>
+  </h1>
+  <div data-type="metadata" style="display: none;">
+    <h1 data-type="document-title" itemprop="name"><%= I18n.t(:eoc_answer_key_title) %></h1>
+    <%= @metadata.paste %>
+  </div>
+</div>
```
kswanson33, line 9: The name `eob` refers to many different things, so I think this name is going to get confusing down the line. Mind renaming this template to `solutions-container` or `answer-key-wrapper` or something like that?

kswanson33 (changes requested):
1. A few picky change requests about names 😅 
2. I notice that there are a few changes in here that have to do with notes. Anything with `data=type="note"` tends to be book-specific. I think it's best to remove/refactor references to calculus-specific note types (`checkpoint`, `theorem`) before merge.

lib/kitchen/directions/bake_notes.rb
jpslav: With book-specific translations the `Media` title should move into the `/books/calculus/locales/en.yml` translation for the `media-2` key.

lib/kitchen/directions/bake_page_abstracts.rb
```diff
@@ -9,6 +9,21 @@ def self.v1(chapter:)
           abstract.prepend(child: "<h3 data-type='title'>#{I18n.t(:learning_objectives)}</h3>")
         end
       end
+
+      def self.v2(chapter:)
+        chapter.abstracts.each do |abstract|
+          ul = abstract.first('ul')
```
jpslav, line 15: Recommend `first!` b/c if the `ul` child is not found you want it to explode with a meaningful message, otherwise you'll get a weird error about `ul` being `nil` on the next line.

lib/kitchen/directions/bake_page_title.rb
```diff
@@ -0,0 +1,22 @@
+# frozen_string_literal: true
+
+module Kitchen
+  module Directions
+    module BakePageTitle
+      def self.v1(page:, number:)
+        page.search("div[data-type='description']").each(&:trash)
+        page.add_class('chapter-content-module')
```
jpslav, line 8: I don't think these things are really related to "baking a page title"... maybe I suggested calling this the wrong name.  Maybe it is `BakeNonIntroductionPage` or something

lib/kitchen/directions/bake_theorem/v1.rb
```diff
@@ -0,0 +1,28 @@
...
+
+      note_body.prepend(sibling:
+        <<~HTML
+          <div class="os-title">
+            <span class="os-title-label">#{I18n.t(:theorem)} </span>
+            <span class="os-number">#{number}</span>
+            <span class="os-divider"> </span>
+          </div>
+        HTML
+      )
```
jpslav, line 20: make a card/issue to change occurrences of this to `PrependNumberedTitle.v1(in: note_body, title: :theorem, number: number)`

lib/kitchen/directions/bake_toc.rb
```diff
@@ -59,6 +58,12 @@ def self.li_for_composite_chapter(chapter)
 
       def self.li_for_chapter(chapter)
         pages = chapter.element_children.only(PageElement, CompositePageElement)
+        inner_composite_chapter = chapter.element_children.map do |element|
+          case element
+          when CompositeChapterElement
+            li_for_composite_chapter(element)
+          end
+        end.compact.join("\n")
```
jpslav, line 66: To make this follow the existing approach like for `pages`, change this block of added code to

```ruby
inner_composite_chapters = chapter.element_children.only(CompositeChapterElement)
```

and then below line 77 becomes

```ruby
${inner_composite_chapters.map { |composite_chapter| li_for_composite_chapter(composite_chapter) }.join("\n")}
```

lib/kitchen/directions/bake_toc.rb
```diff
@@ -44,7 +44,6 @@ def self.li_for_unit(unit)
 
       def self.li_for_composite_chapter(chapter)
         pages = chapter.element_children.only(CompositePageElement)
-
```
jpslav, line 47: I kind of like leaving the empty line if rubocop doesn't complain :-)

lib/locales/en.yml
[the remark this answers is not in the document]
jpslav: @KarinaMendez2 can you use book-specific translation for this?

jpslav (changes requested):

lib/kitchen/directions/bake_checkpoint.rb
```diff
@@ -0,0 +1,51 @@
...
+
+        solution.replace_children(with:
+          <<~HTML
+            <span class="os-divider"> </span>
+            <a class="os-number" href="##{exercise.id}">#{number}</a>
+            <div class="os-solution-container">#{solution.children}</div>
+          HTML
+        )
+
+        exercise.add_class('unnumbered')
```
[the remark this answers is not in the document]
KarinaMendez2: not related but the order of the classnames in easy-baked version is `os-hasSolution unnumbered` so I moved this line to the end

jpslav (approved):

PR #141, opened 2021-03-01
Figures (selectors)

LostMojo:
Here's one way to handle the fact that statistics uses both 'figcaption' and 'div data-type='title'' to represent a figure caption depending on whether it's on a page or a chapter.

Also added I18n


lib/kitchen/figure_element.rb
```diff
@@ -20,7 +20,25 @@ def initialize(node:, document: nil)
     # @return [Element, nil]
     #
     def caption
-      first('figcaption')
+      first('figcaption', "div[data-type='description']")
```
jpslav, line 23: Can you add to `Kitchen::Selectors::Base` and `Kitchen::Selectors::Standard1` then say 

```ruby
first(selectors.figure_caption)
```

this will allow it to be overridden later if needed

lib/kitchen/figure_element.rb
```diff
@@ -20,7 +20,25 @@ def initialize(node:, document: nil)
...
+    end
+
+    # Returns true if the figure is a child of another figure
+    #
+    # @return [Boolean]
+    #
+
```
jpslav, line 30: I think generally no space here; the last blank line with a `#` can be right up against the method, ditto below

lib/kitchen/figure_element.rb
```diff
@@ -20,7 +20,25 @@ def initialize(node:, document: nil)
...
+    def is_subfigure?
+      parent.node.to_s.end_with?('</figure>')
+    end
+
+    # Increments the count of this element in all of this element's ancestors
+    # unless it's a subfigure
+
```
jpslav, line 37: Add `#` on the blank line here

lib/kitchen/figure_element.rb
```diff
@@ -20,7 +20,25 @@ def initialize(node:, document: nil)
...
+    def count_as_descendant
+      @ancestors.each_pair do |type, ancestor|
+        @counts_in[type] = ancestor.increment_descendant_count(short_type) unless is_subfigure?
```
jpslav, line 40: cool solution! much better than my idea.

here's a tweak:  instead of copying this whole method from `ElementBase`, modify the method in `ElementBase` to call another method to see if should be counted.  

```ruby
# element_base.rb
def count_as_descendant
  @ancestors.each_pair do |type, ancestor|
    @counts_in[type] = ancestor.increment_descendant_count(short_type) if countable?
  end
end

def countable?
  true
end
```

With this approach, subclasses can opt out of counting by overriding `countable?` instead of the whole `each_pair` loop, e.g. `FigureElement` can have:

```ruby
def countable?
  !subfigure?
end
```

Also an optimization to the `ElementBase` approach is to put the check first, because there's no need to do all of the iterating if `countable?` is false:

```ruby
# element_base.rb
def count_as_descendant
  return unless countable?

  @ancestors.each_pair do |type, ancestor|
    @counts_in[type] = ancestor.increment_descendant_count(short_type)
  end
end
```

You'll need to add to the element_base_spec.rb to test that `countable?` is used in counting.

lib/kitchen/figure_element.rb
```diff
@@ -20,7 +20,25 @@ def initialize(node:, document: nil)
...
+      first('figcaption', "div[data-type='description']")
+    end
+
+    # Returns true if the figure is a child of another figure
+    #
+    # @return [Boolean]
+    #
+
+    def is_subfigure?
+      parent.node.to_s.end_with?('</figure>')
```
jpslav, line 32: does `parent.name == 'figure'` work?

I know this is covered by `BakeFigure`'s spec, but might be nice to add a simple spec to `figure_element_spec.rb` to test this directly.

lib/kitchen/figure_element.rb
jpslav: Also, while this is test covered by `BakeFigure`'s usage of it, might be nice to have a quick spec in the `figure_element_spec.rb` to show that this captures both kinds of captions.

jpslav (changes requested):

spec/directions/bake_figure_spec.rb
```diff
@@ -42,6 +62,7 @@
               <span class="os-title-label">Figure </span>
               <span class="os-number">1.2</span>
               <span class="os-divider"> </span>
+              <span class="os-title" data-type="title" id="">This Is A Title</span>
```
LostMojo, line 65: So it turns out that data-type="title" and figcaption represent _very **very** similar_ things, but not the _same_ thing. I fixed that and fixed figures & spec to match.

spec/directions/bake_figure_spec.rb
```diff
@@ -76,4 +106,14 @@
     end
   end
 
+  describe '#subfigure?' do
+    it 'can detect subfigures' do
+      is_subfigure = []
+      book2.chapters.figures.each do |figure|
+        is_subfigure.push(figure.subfigure?)
+      end
+      expect(is_subfigure).to eq([false, true, true])
```
LostMojo, line 115: Is this an okay way to test this?

jpslav: Yes this is reasonable as part of the testing, but should collapse it to

```ruby
expect(book2.chapters.figures.map(&:subfigure?)).to eq([false, true, true])
```

`.map` does a `.each` accumulating results into an array for you.  Remember that `.map(&:subfigure?)` is Ruby shorthand for `.map{|figure| an_array_that_gets_returned.push(figure.subfigure?)}` which is shorthand for what you have above.

could even take out `.chapters` 

**My other question is, where do we test that recipe code like**

```ruby
chapter.figures.each do |figure|
  BakeFigure.v1(figure: figure,
                number: "#{chapter.count_in(:book)}.#{figure.count_in(:chapter)}")
end
```

is doing the right thing? like what is `count_in(:chapter)` for a subfigure?

spec/directions/bake_figure_spec.rb
jpslav: Thank you for this comment.

lib/kitchen/directions/bake_figure.rb
```diff
@@ -6,15 +6,17 @@ module BakeFigure
...
 
         caption = figure.caption&.cut
         figure.append(sibling:
           <<~HTML
             <div class="os-caption-container">
-              <span class="os-title-label">Figure </span>
+              <span class="os-title-label">#{I18n.t(:figure)} </span>
               <span class="os-number">#{number}</span>
               <span class="os-divider"> </span>
+              #{"<span class=\"os-title\" data-type=\"title\" id=\"#{title.id}\">#{title.children}</span>" if title}
```
jpslav, line 19: Interestingly, this explains why there are two consecutive `<span class="os-divider"> </span>` in baked content :-)

spec/element_base_spec.rb
```diff
@@ -66,6 +66,12 @@
     end
   end
 
+  describe '#countable?' do
+    it 'returns false if not countable (default true)' do
```
jpslav, line 70: spec description is kind of confusing.  `it 'defaults to true'` would be clearer

jpslav (changes requested):

jpslav:
Also @LostMojo this PR needs a changelog entry.  Dunno why the changelog check is passing.

spec/directions/bake_figure_spec.rb
LostMojo: Just realized, should FigureElement method specs belong to a figure_element_spec file? We don't have one yet.

jpslav (approved):

PR #146, opened 2021-03-02
Bake chapter key concepts

KarinaMendez2:
for: https://app.zenhub.com/workspaces/ce-styles-5cdc84baee941c07853fbbf2/issues/openstax/kitchen/139

wrote this v1 for calculus


lib/kitchen/directions/bake_chapter_key_concepts.rb
```diff
@@ -0,0 +1,57 @@
...
+        fix_titles = chapter.pages.map do |page|
+          key_concepts = page.search('section.key-concepts')
+          next if key_concepts.none?
+
+          key_concepts.search('h3').trash
+          key_concepts.map do |key_concept|
+            id = key_concept.id.split('fs-id')
+            key_concept.prepend(child:
+              <<~HTML
+                <a href="##{id[0]}0">
```
jpslav, line 20: I wonder if this is always supposed to end in `0` and if so what it links to?  Seems a little weird.

lib/kitchen/directions/bake_chapter_key_concepts.rb
```diff
@@ -0,0 +1,57 @@
...
+                </a>
+              HTML
+            )
+
+            <<~HTML
+              <div class="os-section-area">
+                #{key_concept}
+              </div>
+            HTML
+          end.compact.join("\n")
```
jpslav, line 35: Do you need the `compact`?  Would only be useful if `key_concept` were nil (`compact` gets rid of `nil` values in arrays).  Ditto for next line.

spec/directions/bake_chapter_key_concepts_spec.rb
```diff
@@ -0,0 +1,72 @@
+# frozen_string_literal: true
+
+require 'spec_helper'
+
+RSpec.describe Kitchen::Directions::BakeChapterKeyConcepts do
+  before do
+    stub_locales({
+      'eoc_key_concepts': 'Key Concepts'
```
jpslav, line 8: Can you add `eoc_key_concepts` to the real `en.yml` translation file?

spec/directions/bake_chapter_key_concepts_spec.rb
```diff
@@ -0,0 +1,72 @@
...
+            <div data-type="document-title">Page 1</div>
+            <section id="auto_7c_fs-id1" class="key-concepts">
+              <a href="#auto_7c_0">
+                <h3 data-type="document-title" id="auto_7c_0_copy_XXX">
+                  <span class="os-number">1.2</span>
+                  <span class="os-divider"> </span>
+                  <span class="os-text" data-type="" itemprop="">Page 1</span>
+                </h3>
+              </a>
+              <p>Concepts blah.</p>
```
jpslav, line 42: Can you check that in the baked output this stuff is duplicated here and below?

jpslav (changes requested):
Please add a changelog entry

spec/directions/bake_chapter_key_concepts_spec.rb
KarinaMendez2: it's added but on the `bake exercises` branch

spec/directions/bake_chapter_key_concepts_spec.rb
KarinaMendez2: yeah it looks like this because it does `chapter append` not `replace`

spec/directions/bake_chapter_key_concepts_spec.rb
jpslav: ok, we can't merge this one until that is in, and it is currently not as far along as this one?

KarinaMendez2: I'll add the change here and remove it from there, that PR might not make it that far

lib/kitchen/directions/bake_chapter_key_concepts.rb
```diff
@@ -0,0 +1,59 @@
...
+          next if key_concepts.none?
+
+          key_concepts.search('h3').trash
+          key_concepts.map do |key_concept|
+            id = key_concept.id.split('fs-id')
+            key_concept.prepend(child:
+              <<~HTML
+                <a href="##{id[0]}0">
+                  <h3 data-type="document-title" id="#{id[0]}0">
+                    <span class="os-number">#{chapter.count_in(:book)}.#{page.count_in(:chapter) - 1}</span>
```
jpslav, line 22: The `- 1` leads to numbers like `1.0` which I see in the spec.  But baked Calculus always has key concepts starting at `x.1`, not `x.0`.  

I think this is because the book you are working on has introduction pages that are uncounted at the start of each chapter.  This code may only be used in books with introductions in each chapter, but it may not be.  

You can [see in the Chemistry recipe that we get around this by only iterating over the non-introduction pages](https://github.com/openstax/recipes/blob/main/books/chemistry/bake#L62-L70), which ensures that our `page.count_in(:chapter)` always starts at 1 for the first non-introductory page.   You can do the same thing up on line 11: `chapter.pages('$:not(.introduction)')` and then get rid of the `- 1`.

The fact that this crops up in at least one place in a recipe and in two other directions leads me to think that we should add a `non_introduction_pages` method to `ElementEnumeratorBase` [here](https://github.com/openstax/kitchen/blob/main/lib/kitchen/element_enumerator_base.rb#L53) and also a delegator method from `ElementBase` [here](https://github.com/openstax/kitchen/blob/main/lib/kitchen/element_base.rb#L597).  Then you can say `chapter.non_introduction_pages` here and the other places.

lib/kitchen/directions/bake_chapter_key_concepts.rb
```diff
@@ -0,0 +1,59 @@
...
+
+            new_key_concept = key_concept.cut
+
+            <<~HTML
+              <div class="os-section-area">
+                #{new_key_concept.paste}
+              </div>
+            HTML
+          end.join("\n")
+        end.join("\n")
```
jpslav, line 38: This way of doing things is ok and is how I started when I first wrote Kitchen.  But because it is a little hard to follow, I added Rails-like ERB processing so you can make a `blah.xhtml.erb` file with embedded ruby [like this one for BakeIndex](https://github.com/openstax/kitchen/blob/main/lib/kitchen/directions/bake_index/v1.xhtml.erb).  Makes the building more complex XHTML easier to read and understand.

lib/kitchen/directions/bake_chapter_key_concepts.rb
```diff
@@ -0,0 +1,59 @@
...
+              HTML
+            )
+
+            new_key_concept = key_concept.cut
+
+            <<~HTML
+              <div class="os-section-area">
+                #{new_key_concept.paste}
+              </div>
+            HTML
```
jpslav, line 36: I think instead of cutting and pasting you can also say `key_concept.wrap(%(<div class="os-section-area">))`

`%(...)` is just a way to define a string without having to escape all of the double quotes as `\"`.

jpslav (changes requested):

lib/kitchen/element_enumerator_base.rb
```diff
@@ -51,6 +51,17 @@ def pages(css_or_xpath=nil)
...
 
+    # Returns an enumerator that iterates through pages that arent the introduction page within the scope of this enumerator
+    #
+    # @param css_or_xpath [String] additional selectors to further narrow the element iterated over;
+    #   a "$" in this argument will be replaced with the default selector for the element being
+    #   iterated over.
+    #
+    def non_introduction_pages
+      block_error_if(block_given?)
+      chain_to(PageElementEnumerator, css_or_xpath: '$:not(.introduction)')
```
jpslav, line 81: this probably needs to be updated for the `search_query` way in `main`.  Means you'd have `def non_introduction_pages(only: nil, except: nil)` too

lib/kitchen/directions/bake_chapter_key_concepts/v1.rb
```diff
@@ -0,0 +1,38 @@
+# frozen_string_literal: true
+
+module Kitchen::Directions::BakeChapterKeyConcepts
+  class V1
+    renderable
+    def bake(chapter:, metadata_source:)
+      metadata_elements = metadata_source.children_to_keep.copy
+      @chapter_metadata = metadata_elements.paste
```
jpslav, line 8: This works and is ok, but you could remove line 8 and have line 7 have the `@` up front and then say `@metadata_elements.paste` in the `.xhtml.erb` file.

jpslav (changes requested):

PR #149, opened 2021-03-12
Restrict enumerator searches with only and except

jpslav:
See README


lib/kitchen/element_enumerator_factory.rb
```diff
@@ -65,38 +61,34 @@ def or_with(other_factory)
...
             node: sub_node,
             document: element.document,
             element_class: sub_element_class,
-            default_short_type: Utils.search_path_to_type(css_or_xpath),
+            default_short_type: search_query.as_type,
             detect_element_class: detect_sub_element_class
           )
 
-          # If the provided `css_or_xpath` has already been counted, we need to uncount
+          next unless search_query.conditions_match?(sub_element)
```
jpslav, line 80: This is the meat of the PR.  Filters out elements based on the `only` and `except` conditions in `search_query`

kswanson33:
I haven't dug into this error fully, but the chemistry bake fails on this branch with `/code/kitchen/lib/kitchen/ancestor.rb:44:in 'decrement_descendant_count': An element cannot have negative descendants (RangeError)`. I'll look into this more on Monday.

jpslav:
I see at least one bug in the code I need to fix, not sure if it is this bug

kswanson33:
Now chemistry bakes just fine 👍

lib/kitchen/element_base.rb
```diff
@@ -245,67 +245,69 @@ def ancestor_elements
...
-    end
-
-    # Track the sub elements with the given selectors have been counted
-    #
-    # @param css_or_xpath [String] the selectors matching elements that need to be counted
-    # @param count [Integer] The count of those matching subelements
-    #
-    def remember_that_sub_elements_are_already_counted(css_or_xpath:, count:)
-      @css_or_xpath_that_has_been_counted[css_or_xpath] = count
+      @ancestors[ancestor_type].get_descendant_count(short_type)
```
jpslav, line 253: Add `|| raise("No ancestor of type '#{ancestor_type}'")`

kswanson33 (approved):

PR #152, opened 2021-03-15
bake unit titles

brittany-johnson:
relating to [#3506](https://github.com/openstax/cnx-recipes/issues/3506)

### DIFF
--- a/spec/directions/bake_unit_title/v1_spec.rb
+++ b/spec/directions/bake_unit_title/v1_spec.rb
@@ -1,0 +1,43 @@
+# frozen_string_literal: true
+
+require 'spec_helper'
+
+RSpec.describe Kitchen::Directions::BakeUnitTitle::V1 do
+  let(:book1) do
+    book_containing(html:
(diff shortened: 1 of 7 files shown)


spec/directions/bake_unit_title/v1_spec.rb
```diff
@@ -0,0 +1,43 @@
...
+
+RSpec.describe Kitchen::Directions::BakeUnitTitle::V1 do
+  let(:book1) do
+    book_containing(html:
+      <<~HTML
+        <div data-type="unit">
+          <h1 data-type="document-title">Title</h1>
+        </div>
+        <div data-type="unit">
+          <h1 data-type="document-title">Title</h1>
```
jpslav, line 13: Can you change this title to something else to test that it isn't accidentally reusing old unit titles? (I know it isn't in this case, but would be a better test if the string here was different)

jpslav (changes requested):

jpslav (approved):

PR #154, opened 2021-03-16
Title and optional always captions for tables

LostMojo:
Adds a v2 for styling numbered tables in stats.

 The differences include:
* Adds a space to the end of the os-table class (to match original)
* Makes sure there's a caption span in every caption-container, whether there's a caption or no
* adds a os-title span if one can be found within the caption
* adds the text from the os-title span and the caption to the summary


lib/kitchen/directions/bake_numbered_table/v2.rb
```diff
@@ -0,0 +1,63 @@
+# frozen_string_literal: true
+
+module Kitchen::Directions::BakeNumberedTable
+  class V2
+    def bake(table:, number:)
+      # The differences include:
+      # Adding a space to the end of the os-table class (to match original)
+      # Makes sure there's a caption span in every caption-container, whether there's a caption or no
+      # adds a possibility of a os-title span if one can be found within the caption
+      # adds the text from the os-title span
```
LostMojo, line 10: Is this the best place to note differences?

LostMojo:
I forgot specs, one moment

lib/kitchen/directions/bake_numbered_table/v2.rb
jpslav: Maybe more in `main.rb` so someone looking at v1 and v2 could tell the diff?

jpslav:
Talked to @LostMojo separately about seeing if v2 can be part of v1 with a few configurable controls

lib/kitchen/directions/bake_numbered_table/v1.rb
```diff
@@ -22,14 +22,28 @@ def bake(table:, number:)
       table.parent.add_class('os-column-header-container') if table.column_header?
 
       # TODO: extra spaces added here to match legacy implementation, but probably not meaningful?
-      new_summary = "#{table_label}  "
+      new_summary = "#{table_label} "
       new_caption = ''
+      caption_title = ''
+
+      if (title = table.first("span[data-type='title']")&.cut)
```
jpslav, line 29: I think it is time for a `titles` method because we do this search a lot.  After this PR, can you make another kitchen PR that adds `ElementEnumeratorBase#titles` and `ElementBase#titles`.  @KarinaMendez2 just did something similar for metadata searches in https://github.com/openstax/kitchen/pull/133.

jpslav (approved):
Approved - see my one request for a new PR

PR #157, opened 2021-03-22
Fix repeating key concepts

kswanson33:


spec/directions/bake_chapter_key_concepts_spec.rb
```diff
@@ -5,15 +5,22 @@
 RSpec.describe Kitchen::Directions::BakeChapterKeyConcepts do
   let(:chapter) do
     chapter_element(
-      page_element(
-        <<~HTML
+      <<~HTML
+        <div data-type="page">
```
KarinaMendez2, line 9: is page_element not working for you? I think it's better to use it below as well

KarinaMendez2 (changes requested):
works great! one small comment

KarinaMendez2 (approved):

PR #159, opened 2021-03-23
adds titles

LostMojo:
to ElementBase and ElementEnumerator

### DIFF
--- a/lib/kitchen/element_enumerator_base.rb
+++ b/lib/kitchen/element_enumerator_base.rb
@@ -177,6 +177,20 @@ module Kitchen
       chain_to(ExampleElementEnumerator, css_or_xpath: css_or_xpath, only: only, except: except)
     end
 
+    # Returns an enumerator that iterates through titles within the scope of this enumerator
+    #
+    # @param only [Symbol, Callable] the name of a method to call on an element or a
+    #   lambda or proc that accepts an element; elements will only be included in the
+    #   search results if the method or callable returns true
+    # @param except [Symbol, Callable] the name of a method to call on an element or a
+    #   lambda or proc that accepts an element; elements will not be included in the
+    #   search results if the method or callable returns false
+    #
+    def titles(only: nil, except: nil)
+      block_error_if(block_given?)
+      chain_to(ElementEnumerator, css_or_xpath: '[data-type="title]"', only: only, except: except)
+    end
+
     # Returns an enumerator that iterates through metadata within the scope of this enumerator
     #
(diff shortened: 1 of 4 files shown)


lib/kitchen/element_enumerator_base.rb
```diff
@@ -177,6 +177,20 @@ def examples(css_or_xpath=nil, only: nil, except: nil)
...
+    # @param only [Symbol, Callable] the name of a method to call on an element or a
+    #   lambda or proc that accepts an element; elements will only be included in the
+    #   search results if the method or callable returns true
+    # @param except [Symbol, Callable] the name of a method to call on an element or a
+    #   lambda or proc that accepts an element; elements will not be included in the
+    #   search results if the method or callable returns false
+    #
+    def titles(only: nil, except: nil)
+      block_error_if(block_given?)
+      chain_to(ElementEnumerator, css_or_xpath: '[data-type="title]"', only: only, except: except)
```
jpslav, line 191: ok, this is an advanced thing that would not be obvious

You know how we can call the `examples` enumerator method above like `some_element.examples('$.checkpoint')` and it will only iterate over elements that have `[data-type="example"]` and class `checkpoint`?  That happens because `ExampleElementEnumerator` (passed on line 177 above) is the code that knows about `[data-type="example"]` and the `css_or_xpath` that is passed in on line 175 is the variable that takes things like `'$.checkpoint`.  Inside the enumerator building code we merge these two CSS selectors, substituting the `[data-type="example"]` into the `$` to get `[data-type="example"].checkpoint` as our overall selector.

Here it isn't as easy to do that b/c `ElementEnumerator` doesn't have a default css to work with.  But we can fake it.  Actually it is hard to write up, I may make a PR on your PR.

LostMojo:
Getting the enumerator specs-- a little curious, the other enumerators in element_enumerator_base don't seem to have specs?

spec/element_base_spec.rb
```diff
@@ -293,6 +305,12 @@
     end
   end
 
+  describe '#titles' do
+    it 'returns elements with data-type title' do
+      expect(title_book.titles.to_s).to eq('<div data-type="title">This is a title</div>')
```
jpslav, line 310: Can you put two titles in the `title_book`, one that is a `div` and one that is a `span`?  

```
<div data-type="title">One</div>
<span data-type="title">Two</span>
```

And then have your test show that the iterator gets both?  

```ruby
expect(title_book.titles.map(&:text)).to eq(["One", "Two"])
```

jpslav (changes requested):

jpslav (approved):

PR #161, opened 2021-03-24
Note titles

kswanson33:
1. Restructure the format that autogenerated note titles expect from i18n. 
2. Allow recipe to infer or be given book-specific locale files


lib/kitchen/oven.rb
```diff
@@ -12,10 +12,16 @@ class Oven
...
     #
-    def self.bake(input_file:, recipes:, output_file:, config_file: nil)
+    def self.bake(input_file:, recipes:, output_file:, config_file: nil, locales: nil)
       profile = BakeProfile.new
       profile.started!
 
+      if locales
+        I18n.load_path << file_glob("#{__dir__}/#{locales}/*.yml")
+        I18n.backend.reload!
+        # TODO: reload! clears the symbols that are manually stored from TRANSLITERATIONS in openstax_kitchen.rb. Fix: add these to the load_path instead?
```
kswanson33, line 22: @jpslav Concerns I have: 1) the comment, 2) currently in recipes, I am able to pass a locales path that references the locally mounted copy of kitchen. Will this work when not using a local version of kitchen?

jpslav: See my PR into your PR.  It is a different approach that avoids modifying the Kitchen I18n backend.  Also in my PR you'll see that it auto-detects a `locales` directory next to the recipe or you can pass in a path to another locales folder.  The trick when passing in a path is to use `__dir__` to reference the current file's directory.

jpslav: See my PR into your PR.  It is a different approach that avoids modifying the Kitchen I18n backend.  Also in my PR you'll see that it auto-detects a `locales` directory next to the recipe or you can pass in a path to another locales folder.  The trick when passing in a path is to use `__dir__` to reference the current file's directory.

lib/kitchen/note_element.rb
```diff
@@ -62,10 +58,26 @@ def self.is_the_element_class_for?(node)
 
     protected
 
-    def translation_key_in(possible_translation_keys)
-      keys = possible_translation_keys & classes
+    def key
+      if I18n.t(document.short_name).instance_of?(String)
```
jpslav, line 62: Now that we have book-specific translations, I think we can take `chemistry` out of the translation keys don't you think?  B/c if we have `/books/chemistry/locales/en.yml` those translations are for chemistry by definition.  So I think we can get away from lookup by `document.short_name`.  

```yaml
en: 
  note_titles:
    chemist-portrait: Portrait of a Chemist
```

lib/kitchen/note_element.rb
```diff
@@ -62,10 +58,26 @@ def self.is_the_element_class_for?(node)
 
     protected
 
-    def translation_key_in(possible_translation_keys)
-      keys = possible_translation_keys & classes
+    def key
```
jpslav, line 61: `key` is maybe too generic for this method name.  `detected_note_title_key` or something is probably better

lib/kitchen/note_element.rb
```diff
@@ -62,10 +58,26 @@ def self.is_the_element_class_for?(node)
...
+      keys = possible_keys & classes
+
       raise("too many translation keys: #{keys.join(', ')}") if keys.many?
 
+      unless keys.first.present?
+        puts "Title for note with classes #{classes} not found"
+        return
+      end
+
       keys.first
```
jpslav, line 81: Memoize this whole method -- we call it twice, once in `indicates_autogenerated_title?` and then again when we do the actual translation, no need to run all the code twice.

```ruby
def detected_note_title_key
  @detected_note_title_key ||= begin
    ... all the code ending with `keys.first` which is the return value for this block and will store in the var
  end
end
```

lib/kitchen/note_element.rb
```diff
@@ -62,10 +58,26 @@ def self.is_the_element_class_for?(node)
...
 
-    def translation_key_in(possible_translation_keys)
-      keys = possible_translation_keys & classes
+    def key
+      if I18n.t(document.short_name).instance_of?(String)
+        puts "Entry for book with short name #{document.short_name} does not exist in locales"
+        return
+      end
+      unless I18n.t(document.short_name).key?(:notes)
+        puts "The book #{document.short_name} does not have an entry for notes in locales"
```
jpslav, line 67: Can you tell me which of the `puts` and `raise` things here should stay and which should become `RecipeError` or other?

jpslav (changes requested):
Thanks @kswanson33 ! Pls see my PR to this PR

lib/kitchen/note_element.rb
kswanson33: I think we have to keep using the `document.short_name` because of the case where the same class may be in use in multiple books, but mapping to a different title. (ex. I believe `media-2` is used in both u-physics and calculus for different titles.) What do you think?

jpslav: But in such a case those will be different recipe runs so no conflict?

kswanson33: I get it now, works for me 👍

lib/kitchen/note_element.rb
kswanson33: For developers, it seems to me more useful to have `puts`/ a non-fatal response, so that the recipe can still run even if the developer hasn't made a locales file yet. This `puts` feedback also supplies the list of keys that the recipe expects to find in locales, which may be helpful in building the file. But is there a better way of doing this with `raise`/`rescue`?

lib/kitchen/note_element.rb
```diff
@@ -15,6 +15,7 @@ def initialize(node:, document: nil)
             document: document,
             enumerator_class: NoteElementEnumerator,
             short_type: :note)
+      @detected_note_title_key = detected_note_title_key
```
jpslav, line 18: no need for this as `@detected_note_title_key` is only used from within `detected_note_title_key` when it is set.

jpslav (changes requested):

jpslav (approved):

PR #162, opened 2021-03-27
Move recipe-specific locale code into recipe code and use "Chain" I18n backend

jpslav:


lib/kitchen/element_enumerator_base.rb
```diff
@@ -78,7 +78,10 @@ def pages(css_or_xpath=nil, only: nil, except: nil)
     #
     def non_introduction_pages(only: nil, except: nil)
       block_error_if(block_given?)
-      chain_to(PageElementEnumerator, css_or_xpath: '$:not(.introduction)', only: only, except: except)
+      chain_to(PageElementEnumerator,
+               css_or_xpath: '$:not(.introduction)',
+               only: only,
+               except: except)
```
jpslav, line 84: rubocop fix

lib/kitchen/oven.rb
```diff
@@ -12,16 +12,10 @@ class Oven
     # @param recipes [Array<Recipe>] an array of recipes with which to bake the document
     # @param output_file [String] the path to the output file
     #
-    def self.bake(input_file:, recipes:, output_file:, config_file: nil, locales: nil)
+    def self.bake(input_file:, recipes:, output_file:, config_file: nil)
       profile = BakeProfile.new
       profile.started!
 
-      if locales
-        I18n.load_path << file_glob("#{__dir__}/#{locales}/*.yml")
```
jpslav, line 20: I found that load_path stuff was maybe too heavy-handed, as additions to it stick around.  When I looked into it more, I decided to go for a solution where a recipe's translations were added temporarily.  Also the solution in my PR doesn't change the Kitchen backend, it instead makes a new recipe-specific backend and temporarily chains it in front of the standard Kitchen I18n backend.

lib/openstax_kitchen.rb
```diff
@@ -51,7 +51,7 @@ def require_all(relative_folder, file_matcher='**/*.rb')
 
 require_all('kitchen/directions')
 
-I18n.load_path << file_glob('/locales/*.yml')
+I18n.backend.load_translations(file_glob('/locales/*.yml'))
```
jpslav, line 54: Same as above `load_path` comment, `load_path` impacts all backends, so when I made a new temporary backend, it also got all of the kitchen locales loaded onto it.  By calling `load_translations` instead, we avoid that.

spec/oven_spec.rb
```diff
@@ -17,23 +17,64 @@
...
+
+    def expect_bakes(input_xml:, recipes:, output_xml:)
+      input_file = Tempfile.new.tap { |file| file.write(input_xml); file.rewind }
+      output_file = Tempfile.new
+
+      described_class.bake(input_file: input_file.path,
+                           recipes: [recipes].flatten,
+                           output_file: output_file.path)
+
+      expect(output_file.read).to eq output_xml
```
kswanson33, line 77: nice 🖊️📓

kswanson33 (approved):

PR #165, opened 2021-03-31
Add 3 ways of baking notes

kswanson33:
Splits baking notes into 3 versions: `TitledNote` (has a book-specific class corresponding to its title; may or may not have a subtitle), `UnclassifiedNote` (note without a class; may or may not have a title), and `NumberedNote` (similar to `TitledNote`, but numbered). 

The original `BakeNotes` direction hasn't been removed, but `BakeTitledNote` and `BakeUnclassifiedNote` cover everything that `BakeNotes` does.


kswanson33:
@jpslav Questions:
1. At what point should the original `BakeNotes` be removed? Or is it ok to keep it around so that chemistry still works?
2. Is the usage in a recipe I outlined too complicated? If so, is there a way I could make it simpler?
3. I could probably figure out a way to combine `BakeTitledNote` and `BakeUnclassifiedNote`, to follow the way `BakeNotes` worked originally. Is this worth doing, or is it fine the way it is?

lib/kitchen/directions/bake_notes/bake_titled_note.rb
```diff
@@ -0,0 +1,32 @@
...
+
+        return unless title
+
+        title.name = 'h4'
+        title.add_class('os-subtitle')
+        title.replace_children(with:
+          <<~HTML
+            <span class="os-subtitle-label">#{title.children}</span>
+          HTML
+        )
```
jpslav, line 27: `title.wrap_children('span', class: 'os-subtitle-label')`

lib/kitchen/directions/bake_notes/bake_titled_note.rb
```diff
@@ -0,0 +1,32 @@
...
+        note.first!('.os-note-body').prepend(child: title.raw)
```
jpslav, line 28: I saw someone else do this with `title.raw` ... is the `.raw` required?

lib/kitchen/directions/bake_notes/bake_numbered_note.rb
```diff
@@ -0,0 +1,23 @@
+# frozen_string_literal: true
+
+module Kitchen
+  module Directions
+    module BakeNumberedNote
+      def self.v1(note:)
+        note.wrap_children(class: 'os-note-body')
+
+        chapter_count = note.ancestor(:chapter).count_in(:book)
```
jpslav, line 9: why not `note.count_in(:book)`?

jpslav (changes requested):
A minor thing is that `BakeUnclassifiedNote` and `BakeTitledNote` don't require us to iterate through chapters and pages (they don't depend on them), so in the recipe snippet we could do `book.notes.each do |note| ... end` for the top part.

Seems like it'd make sense for `BakeUnclassifiedNote` to know that it only deals with notes without classes (as in that `if note.classes.empty?` line seems like it belongs inside `BakeUnclassifiedNote`.  Would we want to change it `BakeUnclassifiedNotes` (plural) and have its code be like

```ruby
book.notes.each do |note|
  next note.classes.any?
  ... existing code
end
```

Is `BakeTitledNote` really `BakeAutotitledNote`?  Its code seems to assume that the note has an autogenerated title.  As such, since autogenerated titles are derived from classes on the note, should the check for certain classes be something that happens inside this code?  And should it too contain the `book.notes` iteration?

```ruby
module BakeAutotitledNotes
  def self.v1(book:)
    book.notes.each do |note|
      next unless note.indicates_autogenerated_title?
      ... existing code
    end
  end
end
```

Ok and now I see that `BakeNumberedNote` also has autogenerated title stuff... now I'm wondering if indeed these should be separate...

Do you have an example of this working in your real recipe?  That might help shed some light.

lib/kitchen/directions/bake_notes/bake_numbered_note.rb
kswanson33: The numbering of these notes is, for example, 2.1 for the first note in chapter 2. So this isn't counting how many notes are in the book, but instead getting the count of the note's parent chapter.

jpslav: oh duh

lib/kitchen/directions/bake_notes/bake_titled_note.rb
kswanson33: `prepend` seems to want a `Node`, `NodeSet`, or `String`, and reject a `Kitchen::Element`. `to_s` works just as well as `raw`

lib/kitchen/directions/bake_notes/bake_autotitled_note.rb
```diff
@@ -2,8 +2,16 @@
 
 module Kitchen
   module Directions
-    module BakeTitledNote
-      def self.v1(note:)
+    module BakeAutotitledNote
+      def self.v1(book:, selectors:)
```
jpslav, line 6: the second argument can't be any arbitrary selectors because you are comparing against classes.  So I'd call it `classes`

lib/kitchen/directions/bake_notes/bake_numbered_note.rb
```diff
@@ -3,7 +3,15 @@
 module Kitchen
   module Directions
     module BakeNumberedNote
-      def self.v1(note:)
+      def self.v1(book:, selectors:)
```
jpslav, line 6: ditto here on calling it `classes` because `$.#{selector}` assumes it is a class

jpslav (changes requested):

jpslav (approved):

PR #169, opened 2021-04-06
Add ways of baking exercises for uphysics

kswanson33:
Changes in this PR:
- Added `BakeChapterReviewExercises.v2` & specs
- Added a uphysics `Strategy` in `BakeChapterAnswerKey` & specs
- Refactored `BakeChapterReviewExercises.v1` to take a class selector as an argument
- Light refactoring in several files to make metadata more consistent


lib/kitchen/directions/bake_chapter_answer_key/v1.rb
```diff
@@ -17,7 +19,10 @@ def bake(chapter:, metadata_source:, strategy:, append_to:)
             <h2 data-type="document-title">
               <span class="os-text">#{I18n.t(:chapter)} #{chapter.count_in(:book)}</span>
             </h2>
-            #{metadata_source.copy}
+            <div data-type="metadata" style="display: none;">
+              <h1 data-type="document-title" itemprop="name">#{I18n.t(:chapter)} #{chapter.count_in(:book)}</h1>
+              #{metadata_source.children_to_keep.copy.paste}
```
kswanson33, line 24: Just as a note, copying & pasting is what adds the `_copy_XXX` suffix (with `XXX` replaced by generated numbers). This allows for us to make sure that all ids are unique.

spec/directions/bake_chapter_review_exercises/v1_spec.rb
```diff
@@ -47,34 +50,8 @@
...
-          </p>
-        </div>
-        <div itemprop="about" data-type="subject">Book Title</div>
-      </div>
-    HTML
-  end
-
   it 'works' do
-    described_class.new.bake(chapter: book_with_review_exercises.chapters.first, metadata_source: metadata, append_to: append_to)
+    described_class.new.bake(chapter: book_with_review_exercises.chapters.first, metadata_source: metadata_element, append_to: append_to, klass: 'review-exercises')
```
kswanson33, line 54: `metadata_element` is a factory helpers method that returns some generic metadata

lib/kitchen/directions/bake_chapter_review_exercises/v1.rb
```diff
@@ -4,12 +4,16 @@ module Kitchen::Directions::BakeChapterReviewExercises
...
     renderable
 
-    def bake(chapter:, metadata_source:, append_to:)
-      @metadata = metadata_source
-      @exercise_clipboard = Kitchen::Clipboard.new
+    def bake(chapter:, metadata_source:, append_to:, klass:)
+      @klass = klass
+      @metadata = metadata_source.children_to_keep.copy
+      @title = I18n.t(:"eoc.#{klass}")
+      @content = ''
```
KarinaMendez2, line 11: do you set this variable anywhere else? if not, line 34 is enough and you can remove this

lib/kitchen/directions/bake_chapter_answer_key/strategies/uphysics.rb
```diff
@@ -0,0 +1,49 @@
+# frozen_string_literal: true
+
+module Kitchen::Directions::BakeChapterAnswerKey
+  module Strategies
+    class UPhysics
+      def bake(chapter:, append_to:)
+        classes = %w[review-conceptual-questions review-problems review-additional-problems
+                     review-challenge]
+        classes.each do |klass|
+          bake_section(chapter: chapter, append_to: append_to, klass: klass)
```
KarinaMendez2, line 10: is there a reason for the k in klass? maybe with can it selector or something more specific

lib/kitchen/directions/bake_chapter_review_exercises/v2.rb
```diff
@@ -0,0 +1,61 @@
...
+  # Main difference from v1 is the presence of a section title
+  # and some additional wrappers
+  class V2
+    renderable
+
+    def bake(chapter:, metadata_source:, append_to:, klass:)
+      @klass = klass
+      @metadata = metadata_source.children_to_keep.copy
+      @title = I18n.t(:"eoc.#{klass}")
+      @content = ''
```
KarinaMendez2, line 13: same here

KarinaMendez2 (changes requested):

lib/kitchen/directions/bake_chapter_answer_key/strategies/uphysics.rb
kswanson33: oh it's bc `class` is already a thing in Ruby. I do like selector, except for that it seems to imply we could use an xpath as well. Using `klass` is a thing I borrowed from JP. Maybe `tag`?

lib/kitchen/directions/bake_chapter_review_exercises/v1.rb
kswanson33: ohhhhh awesome thanks

KarinaMendez2 (approved):

lib/kitchen/directions/bake_chapter_answer_key/strategies/uphysics.rb
brittany-johnson: `tag` makes me think its referring to the html tag. 
I think `klass` is fine here?

brittany-johnson (approved):

PR #174, opened 2021-04-08
bake exercises in notes for uphysics

kswanson33:
Changes:
- `BakeNumberedNotes` now checks if a note has an exercise; if so, it bakes the exercise and solution in place
- Updated UPhysics strategy in `BakeChapterAnswerKey` to also bake solutions from notes


lib/kitchen/directions/bake_notes/bake_numbered_notes.rb
```diff
@@ -7,6 +7,9 @@ def self.v1(book:, classes:)
         classes.each do |c|
           book.chapters.notes("$.#{c}").each do |note|
             bake_note(note: note)
+            next unless note.exercises.count.positive?
```
KarinaMendez2, line 10: I think `.positive?` returns `true` on 0

kswanson33: image
had to double-check -- it is `false` on 0!

KarinaMendez2: ok awesome! lol

jpslav: Kudos to y'all for checking the behavior of `positive?` 👏 

This could instead be 

```ruby
next unless note.exercises.any?
``` 

or

```ruby
next if note.exercises.none?
```

... but I think it is cleaner if this check is inside `bake_note_exercise`.  That code already has conditionals to return if there isn't a solution, so it seems reasonable to me that it could also be the place that bails out if there is no exercise.  

```ruby
def self.bake_note_exercise(note:)
  exercise = note.exercises.first
  return unless exercise
```

lib/kitchen/directions/bake_notes/bake_numbered_notes.rb
```diff
@@ -26,6 +29,28 @@ def self.bake_note(note:)
           HTML
         )
       end
+
+      def self.bake_note_exercise(note:)
+        exercise = note.exercises.first
+        exercise.solution ? exercise.add_class('os-hasSolution unnumbered') : exercise.add_class('unnumbered')
```
jpslav, line 35: This has behavior that is based on `exercise.solution` being truthy -- this check is duplicated by the `return unless exercise.solution` on line 39.  Also this line always adds the class 'unnumbered' which isn't all that clear right away.  I think it'd be more straightforward to split this into two lines, this first one should be

```ruby
exercise.add_class('unnumbered')
```

to make it clear that this will always happen.  And then after line 39 you can say

```ruby
exercise.add_class('os-hasSolution')
```

lib/kitchen/directions/bake_notes/bake_numbered_notes.rb
```diff
@@ -26,6 +29,28 @@ def self.bake_note(note:)
...
+        # bake problem
+        exercise.problem.wrap_children('div', class: 'os-problem-container')
+        exercise.problem.search('strong').first&.trash
+        return unless exercise.solution
+
+        # bake solution in place
+        exercise.solution[:id] = "#{exercise[:id]}-solution"
+        solution_number = note.search('.os-number').first.children.first
```
jpslav, line 43: Can this be `note.first('.os-number').text`?

lib/kitchen/directions/bake_notes/bake_numbered_notes.rb
```diff
@@ -26,6 +29,28 @@ def self.bake_note(note:)
           HTML
         )
       end
+
+      def self.bake_note_exercise(note:)
+        exercise = note.exercises.first
+        exercise.solution ? exercise.add_class('os-hasSolution unnumbered') : exercise.add_class('unnumbered')
+        # bake problem
+        exercise.problem.wrap_children('div', class: 'os-problem-container')
+        exercise.problem.search('strong').first&.trash
```
jpslav, line 38: `search('strong').first` is the same as `first('strong')` I believe.

lib/kitchen/directions/bake_notes/bake_numbered_notes.rb
```diff
@@ -26,6 +29,28 @@ def self.bake_note(note:)
...
+        return unless exercise.solution
```
jpslav, line 39: We call `exercise.solution` here and then 3 more times below.  Let's memoize it.  Until we merge in my "little bit of caching" PR, every time we call `.solution` it does a new search for it.  

So either

```ruby
solution = exercise.solution
return unless solution
```

or

```ruby
return unless (solution = exercise.solution)
```

and then replace `exercise.solution` with `solution`.  Also makes the code a little more compact

lib/kitchen/directions/bake_notes/bake_numbered_notes.rb
```diff
@@ -7,6 +7,9 @@ def self.v1(book:, classes:)
         classes.each do |c|
```
jpslav, line 7: Let's change the `c` to `klass` -- fyi @KarinaMendez2 I told @kswanson33 I'm not a fan of single letter variable names -- they are hard to search for and hard to understand when reading code that uses them ("what's 'c' in this line?")

jpslav (changes requested):

lib/kitchen/directions/bake_notes/bake_numbered_notes.rb
kswanson33: This is the way it is because the easybaked version has `os-hasSolution` before `unnumbered` in the class ordering, and it shows up in the diff if I don't add these classes in that order. However, now that you say this, I realize that it would probably be better to alphabetize class names in the normalize script instead of making Kitchen deal with it.

kswanson33: I've now made this change https://github.com/openstax/recipes/pull/34

KarinaMendez2 (approved):

jpslav (approved):

Issue #180, opened 2021-04-14
Create v3 for BakeChapterReviewExercises for Sociology recipe.


ajluk:
Related card: https://github.com/openstax/cnx-recipes/issues/3504

- create v3 of `MoveExercisesToEOC`
- adjust `MoveExercisesToEOC` , `BookAnswerKeyContainer` and `MoveSolutionsToAnswerKey` to accept another class (`solution`), the default class has been left as `solutions`
- suppress even solutions in the Answer Key
- fix `BakeFurtherResearch` baking with main `bake` script error
- implement hacky solution for fixing issue with empty titles and container appearing in the Answer Key in Sociology 2e, when the solutions are commented out, so both Sociologies 2e and 3e could use the same recipe.

ajluk:
Closing since https://github.com/openstax/kitchen/pull/181has been merged.

PR #181, opened 2021-04-14
create v3 for MoveExercisesToEOC

ajluk:
Issue: #180


lib/kitchen/directions/bake_further_research.rb
```diff
@@ -12,6 +12,8 @@ def self.v1(chapter:, metadata_source:)
 
         chapter.non_introduction_pages.each do |page|
           further_research = page.first('.further-research')
+          next unless further_research.present?
```
ajluk, line 15: I have added this line, to prevent the error I've got when I baked sociology using the general `bake script  (not `shorten`): 
image

hack empty answer key sociology/if statement in the bake.rb sociology file
```diff
@@ -0,0 +1,9 @@
+if book.search('.os-eob.os-solution-container[data-type = "composite-chapter"]').first.contains?(' > [data-type="composite-page"] > div[data-type="solution"]')
+    book.search('.os-eob.os-solution-container[data-type = "composite-chapter"]').first.add_class('has-solutions')
+    book.search('body').first.add_class('has-answerkey')
+  else
+    book.search('.os-eob.os-solution-container[data-type = "composite-chapter"]').first.add_class('no-solutions')
+    book.search('.os-eob.os-solution-container.no-solutions[data-type = "composite-chapter"]').first.trash
+    book.search('body').first.add_class('no-answerkey')
+    book.search('body.no-answerkey > #toc > ol > li.os-toc-composite-chapter').trash
+  end
```
ajluk, line 9: I have written some similar hacky solution to adjust the sociology recipe so it can be used by both 2e and 3e editions as BrritanyJ have added within this PR in cnx-recipes repo: https://github.com/openstax/cnx-recipes/pull/3829. @brittany-johnson please let me know what are your thougts about writing it this way in Kitchen and if we need something less hacky and which can be reused in other books I can write some direction/module based on it. cc @jpslav

lib/kitchen/directions/bake_numbered_exercise/v1.rb
```diff
@@ -2,15 +2,20 @@
...
 
       if solution.present?
         if suppress_solution
           solution.trash
+        elsif suppress_even_solution && number.odd?
+          problem_number = "<a class='os-number' href='##{exercise.id}-solution'>#{number}</a>"
+          bake_solution(exercise: exercise, number: number)
+        elsif suppress_even_solution && number.even?
+          problem_number = "<span class='os-number'>#{number}</span>"
+          bake_even_solution(exercise: exercise)
```
ajluk, line 18: In Sociology only odd solutions are needed in Answer Key, so I suppressed even ones based on the number of the exercises - it was needed to be done in this file because in that case the link pointing to solution can be only present in the in the odd exercises numbers.

lib/kitchen/directions/book_answer_key_container/main.rb
```diff
@@ -3,8 +3,8 @@
 module Kitchen
   module Directions
     module BookAnswerKeyContainer
-      def self.v1(book:)
-        V1.new.bake(book: book)
+      def self.v1(book:, klass: 'solutions')
+        V1.new.bake(book: book, klass: klass)
```
ajluk, line 7: In Sociology class used in the Answer Key differs (it is `solution`, not solutions), I have left `solutions` as the default value, so it can be still used by other books which have written answer key strategy without breaking the baking process

lib/kitchen/directions/book_answer_key_container/v1.rb
```diff
@@ -3,11 +3,11 @@
 module Kitchen::Directions::BookAnswerKeyContainer
   class V1
     renderable
-
-    def bake(book:)
+    def bake(book:, klass: 'solutions')
```
ajluk, line 6: See above.

lib/kitchen/directions/book_answer_key_container/v1.rb
```diff
@@ -3,11 +3,11 @@
...
+      @klass = klass
       @metadata = book.metadata.children_to_keep.copy
-      book.body.append(child: render(file: 'eob_solutions_container.xhtml.erb'))
-      book.body.first('.os-eob.os-solutions-container')
+      book.body.append(child: render(file: 'eob_answer_key_outer_container.xhtml.erb'))
+      book.body.first(".os-eob.os-#{@klass}-container")
```
ajluk, line 10: See the comment form above.

ajluk:
@jpslav @brittany-johnson I have requested for your reviews precisely because of the hack solution I have implemented in the sociology recipe. Let me know if we can accept it since I guess it is an exceptional case, but if we need something more reusable and solid I will create a direction/module for it. Also I haven't updated the branch because of the problems with baking on current Kitchen version, but also please let me know if I should do that first if you think it is relevant here. Please also tag some other folks if they can/should review this PR. Thank you! :slightly_smiling_face:

spec/directions/bake_numbered_exercise/v1_spec.rb
```diff
@@ -63,4 +63,50 @@
       )
     end
   end
+
+  context 'when even solutions are suppressed' do
+    context 'when number is odd' do
+      it 'works' do
+        described_class.new.bake(exercise: exercise1, number: 1, suppress_even_solution: true)
```
ajluk, line 70: In this context I have changed the number value to 1 (integer) instead of 1.1 (it is recognized as float), because of the error I've got while this test was running: 
image
The methods `odd?, even?` can't be used on such numbers.
I decided to not change the lines I have added from line 13 in `bake_numbered_exercise.rb` file after this error and change the test data instead because the only books that currently utilize suppressing half of the solutions are History and Sociology and there the numbering of the exercises is created to contain only one number (1). These books don't have exercises in notes that are numbered using two counters creating one number (1.1) like Calculus for example. Also the numbers for exercises in notes are created in separate files (f.ex `bake_checkpoint.rb` for checkpoint notes for Calculus recipe) and I wasn't sure If we will ever need to suppress half of the the solutions from exercises in notes. But let me know what do you think about it?

lib/kitchen/directions/bake_further_research.rb
```diff
@@ -24,6 +24,8 @@ def bake(chapter:, metadata_source:, uuid_prefix: '.')
 
           chapter.non_introduction_pages.each do |page|
             further_research = page.first('.further-research')
+            next unless further_research.present?
```
ajluk, line 27: I have added this line, to prevent the error I've got when I baked sociology using the general `bake` script  (not `shorten`): 
image

lib/kitchen/directions/bake_numbered_exercise/v1.rb
jpslav: I made a suggestion in https://github.com/openstax/kitchen/pull/243

lib/kitchen/directions/bake_numbered_exercise/v1.rb
```diff
@@ -43,5 +49,12 @@ def bake_solution(exercise:, number:, divider: '. ')
         HTML
       )
     end
+
+    def bake_even_solution(exercise:)
+      solution = exercise.solution
+      solution.id = "#{exercise.id}-solution"
```
jpslav, line 55: this line isn't needed b/c we trash the solution

lib/kitchen/directions/bake_numbered_exercise/main.rb
```diff
@@ -3,13 +3,17 @@
...
+        V1.new.bake(exercise: exercise, number: number, suppress_solution: suppress_solution, suppress_even_solution: suppress_even_solution)
       end
 
       def self.bake_solution_v1(exercise:, number:, divider: '. ')
         V1.new.bake_solution(exercise: exercise, number: number, divider: divider)
       end
+
+      def self.bake_even_solution_v1(exercise:, divider: '. ')
+        V1.new.bake_even_solution(exercise: exercise, divider: divider)
+      end
```
jpslav, line 16: Is this used?  Now wondering why we have `bake_solution_v1`

spec/directions/bake_numbered_exercise/v1_spec.rb
jpslav: Interested in seeing your recipe to see how I can help here

jpslav (changes requested):

hack empty answer key sociology/if statement in the bake.rb sociology file
brittany-johnson: @ajluk Hmm.. I think instead we should do something different to address this. 

I believe there are two directions for the answer key - one for the container `BookAnswerKeyContainer` and the strategies in `MoveSolutionsToAnswerKey`.

One idea is to add a conditional in sociology recipe to only call those two directions if solutions are found in the book passed to bake. 
I think this could work for both 2e & 3e. 

What do you think? @ajluk

brittany-johnson (changes requested):

hack empty answer key sociology/if statement in the bake.rb sociology file
ajluk: @brittany-johnson Thank you! I have added the changes you suggested in the `recipes` repo to sociology recipe PR directly: https://github.com/openstax/recipes/pull/24. Let me know if this is similar to something you have in mind? :slightly_smiling_face:

lib/kitchen/directions/bake_numbered_exercise/v1.rb
ajluk: @jpslav Thank you for this idea! :slightly_smiling_face: There were some small fixes need in the PR you have created to make sociology recipe be baked with your changes, but it really simplified things.

lib/kitchen/directions/move_exercises_to_eoc/v3.rb
```diff
@@ -0,0 +1,50 @@
...
+      end
+
+      return if exercise_clipboard.none?
+
+      @content = exercise_clipboard.paste
+
+      append_to_element = append_to || chapter
+      @in_composite_chapter = append_to_element.is?(:composite_chapter)
+
+      # append_to_element.append(child: render(file: 'review_exercises.xhtml.erb'))
```
brittany-johnson, line 45: can be deleted?

hack empty answer key sociology/if statement in the bake.rb sociology file
brittany-johnson: Yes! This is what I had in mind - thanks for adding it. 
This file can be deleted now

lib/kitchen/directions/book_answer_key_container/v1.rb
brittany-johnson, line 6: Because you're setting the value of `klass` in main, I think you can remove here and just have `klass:`. 
I was doing a [similar thing in another PR](https://github.com/openstax/kitchen/pull/255/files) and Mary Jo gave me the pointer!

lib/kitchen/directions/book_answer_key_container/eob_answer_key_outer_container.xhtml.erb
```diff
@@ -1,4 +1,4 @@
-<div class="os-eob os-solutions-container" data-type="composite-chapter" data-uuid-key=".solutions">
```
brittany-johnson, line 1: Curious about the renaming of this file?

brittany-johnson (changes requested):
Sorry for taking so long - I have a few questions and comments

lib/kitchen/directions/bake_numbered_exercise/v1.rb
```diff
@@ -12,9 +12,18 @@ def bake(exercise:, number:, suppress_solution: false)
...
+        when Symbol
+          number.send(suppress_solution_if)
+        else
+          suppress_solution_if
+        end
+
       if solution.present?
         if suppress_solution
           solution.trash
+          exercise.add_class('os-hasSolution-trashed') if note_suppressed_solutions
```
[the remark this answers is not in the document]
ajluk: @brittany-johnson No, it is not, I have just left it so we can spot the exercises with removed solutions if we need it for something in the future and to make their class differ form the ones with solutions. What do you think about it?

lib/kitchen/directions/bake_numbered_exercise/v1.rb
```diff
@@ -12,9 +12,18 @@ def bake(exercise:, number:, suppress_solution: false)
       )
       problem_number = "<span class='os-number'>#{number}</span>"
 
+      suppress_solution =
+        case suppress_solution_if
+        when Symbol
+          number.send(suppress_solution_if)
```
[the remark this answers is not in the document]
ajluk: Np! As I understood it is used to pass the :even? method to number as a symbol used in the  bake file: 
image
And when the number of the exercise is even the solution is trashed and exercise gets additional class when note_suppressed_solutions is also true.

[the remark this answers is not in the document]
ajluk: These are differences between how the code looked in this part before and how JP changed it: 
image

lib/kitchen/directions/book_answer_key_container/v1.rb
ajluk: Thank you! I wasn't sure, so I left it there too. :sweat_smile:

lib/kitchen/directions/bake_numbered_exercise/main.rb
```diff
@@ -3,8 +3,9 @@
 module Kitchen
   module Directions
     module BakeNumberedExercise
-      def self.v1(exercise:, number:, suppress_solution: false)
-        V1.new.bake(exercise: exercise, number: number, suppress_solution: suppress_solution)
+      def self.v1(exercise:, number:, suppress_solution_if: false, note_suppressed_solutions: false)
```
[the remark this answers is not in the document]
ajluk: @brittany-johnson I have left it as JP created it, because it can be either true, false or `:even?` passing as a symbol. See my comment here: https://github.com/openstax/kitchen/pull/181/files#r647324906

lib/kitchen/directions/bake_numbered_exercise/v1.rb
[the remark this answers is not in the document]
brittany-johnson: Ahhh that makes sense. Thanks for explaining!

lib/kitchen/directions/bake_numbered_exercise/v1.rb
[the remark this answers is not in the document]
brittany-johnson: ooooo okay thank you!

lib/kitchen/directions/book_answer_key_container/eob_answer_key_outer_container.xhtml.erb
brittany-johnson: ooooohhh that makes sense

brittany-johnson (approved):

ajluk:
The PR needs to be updated with changes from:  #262

jpslav (approved):

jpslav (changes requested):
Please add an entry to CHANGELOG.md

kswanson33 (approved):

jpslav (approved):

PR #182, opened 2021-04-15
Polish translations

wilkus27:
Create translations of strings for Polish Psychology


jpslav (changes requested):
Please add an entry into the Unreleased section of the CHANGELOG.md.

Also, can you have some discussions (after this PR) with the OpenStax US styles team to think about how to tell Kitchen to use Polish vs English translations? cc @KarinaMendez2 @kswanson33 @LostMojo @brittany-johnson

wilkus27:
@jpslav thank you! I added missing entry. 

About using translation, I just added this line for now  to pl-psychology `bake` file in Recipe repo:

image

It's working, but I will bring this up on the call with the team, so we can find more elegant solution :slightly_smiling_face:

jpslav (approved):

PR #183, opened 2021-04-15
Handle ID numbering to ensure unique IDs

kswanson33:


lib/kitchen/document.rb
```diff
@@ -177,6 +177,18 @@ def modified_id_to_paste(original_id)
...
 
+    # Copies an ID, returning a unique ID
+    #
+    # @param id [String]
+    #
+    def copy_id(id)
+      return nil if id.nil?
+      return '' if id.blank?
+
+      @next_paste_count_for_id[id] ? @next_paste_count_for_id[id] += 1 : @next_paste_count_for_id[id] = 1
```
kswanson33, line 188: is there a more syntactically nice way of doing this?

lib/kitchen/document.rb
```diff
@@ -177,6 +177,18 @@ def modified_id_to_paste(original_id)
...
+      "#{id}#{@id_copy_suffix}#{@next_paste_count_for_id[id]}"
+    end
+
```
kswanson33, line 191: To me, this exposes some of the ID incrementing functionality that had previously been nicely encapsulated; however, it is useful in the two modified bake directions. Thoughts?

lib/kitchen/directions/eoc_section_title_link_snippet.rb
```diff
@@ -0,0 +1,22 @@
...
+        copied_id = page.document.copy_id(original_id)
+        <<~HTML
+          <a href="##{original_id}">
+            <h3 data-type="document-title" id="#{copied_id}">
+              <span class="os-number">#{chapter.count_in(:book)}.#{page.count_in(:chapter)}</span>
+              <span class="os-divider"> </span>
+              <span class="os-text" data-type="" itemprop="">#{page.title.text}</span>
+            </h3>
+          </a>
+        HTML
```
jpslav, line 18: Ok, I see now that the problem is that you are not doing a copy/paste but want the behavior of a pasted ID.

First alternative: Can we add a `copied_id` method to `ElementBase` that hides some of the work here?

```ruby
def self.v1(page:)
  chapter = page.ancestor(:chapter)
  <<~HTML
    <a href="##{page.title.id}">
      <h3 data-type="document-title" id="#{page.title.copied_id}">
        <span class="os-number">#{chapter.count_in(:book)}.#{page.count_in(:chapter)}</span>
        <span class="os-divider"> </span>
        <span class="os-text" data-type="" itemprop="">#{page.title.text}</span>
      </h3>
    </a>
  HTML
end
```

Then `copied_id` is either a call to your new methods (or their guts) or would this work?

```ruby
def copied_id
  document.record_id_copied(id)
  document.modified_id_to_paste(id)
end
```

Second alternative: Does this way work without needing the `copy_id`?

```ruby
def self.v1(page:)
  chapter = page.ancestor(:chapter)
  page_title = page.title
  link_title = page_title.copy

  link_title.wrap_children('span', class: 'os-text', data_type: '', itemprop: '')
            .prepend(child: 
              <<~HTML
                <span class="os-number">#{chapter.count_in(:book)}.#{page.count_in(:chapter)}</span>
                <span class="os-divider"> </span>
              HTML
            )

  <<~HTML
    <a href="##{page_title[:id]}">
      #{link_title.paste}
    </a>
  HTML
end
```

lib/kitchen/document.rb
jpslav: May not be necessary per above, but there might be a way to make this better using hash default values

lib/kitchen/document.rb
jpslav: This method may not be necessary, but if it is let's revisit these questions.

jpslav (changes requested):

lib/kitchen/directions/eoc_section_title_link_snippet.rb
kswanson33: Thank you! I like the first alternative. I tried out the second alternative in the process of settling on this solution, and it did work, but I think I felt like it would be too hard to debug and/or express to future developers why it was important that the page title go through the copy/paste process. What do you think?

jpslav: I agree the first alternative is the best.  Makes it clear what is going on with straightforward code.

PR #184, opened 2021-04-15
Bake problems first elements 

KarinaMendez2:


lib/kitchen/directions/bake_problem_first_elements.rb
```diff
@@ -0,0 +1,16 @@
+# frozen_string_literal: true
+
+module Kitchen
+  module Directions
+    module BakeProblemFirstElements
+      def self.v1(within:, selectors:)
+        selectors.each do |selector|
+          within.search("#{selector}:first-child").each do |problem|
```
jpslav, line 8: Can `.os-problem-container > ` be hardcoded here b/c this is all about baking PROBLEM first elements?

jpslav (changes requested):

jpslav (approved):

PR #186, opened 2021-04-19
Very small change to bake calculus notes with exercises

KarinaMendez2:


kswanson33 (changes requested):
Could you add/change the spec to include an example with `data-type="commentary"`?

jpslav (approved):

PR #190, opened 2021-04-20
Allow unit and page title text to be retrieved regardless of bake status

kswanson33:


lib/kitchen/page_element.rb
```diff
@@ -28,6 +28,14 @@ def title
       first!(is_introduction? ? selectors.title_in_introduction_page : selectors.title_in_page)
     end
 
+    # Returns the title's text regardless of whether the title has been baked
+    #
+    # @return [String]
+    #
+    def title_text
+      title.children.one? ? title.text : title.first('.os-text').text
+    end
```
jpslav, line 37: Seems reasonable.  Should this be on `ElementBase`?  Also is there a spec that tests the `title.text` pre-bake behavior?

jpslav (changes requested):

lib/kitchen/page_element.rb
kswanson33: `ElementBase` doesn't currently have a `title` method, so it doesn't seem to make sense to give it a `title_text` method, especially since not all elements have titles. I'm also not sure if all baked titles get the `os-text` class, although it seems like a fair assumption. What do you think? `ElementBase` could have `title` and `title_text` & just return nil if the title doesn't exist.

The sample data in the specs contains both pre and post bake html structures. Both behaviors don't show up in the diff because one already existed, so I only had to change the other to make sure both were covered.

jpslav (approved):

Issue #185, opened 2021-04-17
Detect content locale and set during baking


jpslav:
Our assembled XHTML files include the language of the content in the `<html>` tag, e.g:

```html
<html xmlns="http://www.w3.org/1999/xhtml" xmlns:m="http://www.w3.org/1998/Math/MathML" lang="en">
```

Add functionality to Kitchen to use the `lang` attribute of the `<html>` tag to set the appropriate I18n locale.

Add a `Kitchen::Document#locale` method that returns the `lang` attribute of the `html` tag as a Ruby symbol.  If no such attribute is present, [emit a warning](https://github.com/openstax/kitchen/blob/fb45d09eaa09f271aa30154579581143df152e05/lib/kitchen/directions/bake_exercises/main.rb#L7) like "No content language found in input file, defaulting to `en`" and return the symbol for English, `:en`.

Modify `Kitchen::Oven#bake` to set `I18n.locale = document.locale`.

Make sure to add unit tests.

PR #192, opened 2021-04-23
Bake direct eoc exercises

wilkus27:
In `pl-psychology` EOC exercises sections are direct children of chapter. (They're not inside composite chapter like in `u-physics` or `calculus`). 

I created two `erb` files: one for direct exercises section (`direct_review_exercises.xhtml.erb`) and one for exercises section from composite chapter (`child_review_exercises.xhtml.erb`). They differ only in two places, title tag and title from metadata:

image

Let me know what do you think about this solution and name convention of files.


lib/kitchen/directions/bake_chapter_review_exercises/v1.rb
```diff
@@ -32,7 +32,13 @@ def bake(chapter:, metadata_source:, append_to:, klass:)
 
       @content = exercise_clipboard.paste
 
-      append_to.append(child: render(file: 'review_exercises.xhtml.erb'))
+      append_to_element = append_to || chapter
+
+      if append_to_element == append_to
+        append_to_element.append(child: render(file: 'child_review_exercises.xhtml.erb'))
+      elsif append_to_element == chapter
+        append_to_element.append(child: render(file: 'direct_review_exercises.xhtml.erb'))
```
LostMojo, line 40: Don't forget to write specs for this change!

lib/kitchen/directions/bake_chapter_review_exercises/direct_review_exercises.xhtml.erb
```diff
@@ -0,0 +1,10 @@
+<div class="os-eoc os-<%=@klass%>-container" data-type="composite-page" data-uuid-key=".<%=@klass%>">
+  <h2 data-type="document-title">
+    <span class="os-text"><%= @title %></span>
+  </h2>
+  <div data-type="metadata" style="display: none;">
+    <h1 data-type="document-title" itemprop="name"><%= @title %></h1>
```
LostMojo, line 6: Will this work with internationalization?

LostMojo (changes requested):
You've got an interesting solution! We might be able to cut it down to one base file if we just have a tag for h2/h3. 

Also, don't forget to write spec for your changes!

kswanson33 (commented):
I like this solution as a way of managing the differences between books for the eoc exercises. I agree with @LostMojo -- I think we can use just one template file for both ways. An example would be something like:

```html
<div class="os-eoc os-<%=@klass%>-container" data-type="composite-page" data-uuid-key=".<%=@klass%>">
  <<%= @tag %> data-type="document-title">
    <span class="os-text"><%= @title %></span>
  </<%= @tag %>>
  <div data-type="metadata" style="display: none;">
    <h1 data-type="document-title" itemprop="name"><%= @metadata_title %></h1>
    <%= @metadata.paste %>
  </div>
  <%= @content %>
</div>
```

where `@tag` is either `h2` or `h3` and `@metadata_title` is either `@title` or `I18n.t(:eoc_exercises_title)`

lib/kitchen/directions/bake_chapter_review_exercises/v1.rb
wilkus27: After applying solution suggested by Kendra there is no need for new specs. All tests passed.

lib/kitchen/directions/bake_chapter_review_exercises/direct_review_exercises.xhtml.erb
wilkus27: For now I just added this line to `bake` file:

image

And it's working. In the future I'm going to apply such solution: https://github.com/openstax/kitchen/issues/185.

wilkus27:
@kswanson33 @LostMojo Thank you for all your comments :relaxed: I applied solution suggested by you. It's ready for rereview.

lib/kitchen/directions/bake_chapter_review_exercises/v1.rb
```diff
@@ -32,7 +32,11 @@ def bake(chapter:, metadata_source:, append_to:, klass:)
 
       @content = exercise_clipboard.paste
 
-      append_to.append(child: render(file: 'review_exercises.xhtml.erb'))
+      append_to_element = append_to || chapter
+      @tag = append_to ? 'h3' : 'h2'
+      @metadata_title = append_to ? I18n.t(:eoc_exercises_title) : @title
```
kswanson33, line 37: Even though Codecov says it's covered, I notice that the specs for v1 & v2 only have an example of this method being called with an append_to element, not without, so if there's anything unexpected with the behavior around appending to chapter, the spec won't catch it. Could you add that example to [the specs](https://github.com/openstax/kitchen/tree/main/spec/directions/move_exercises_to_eoc)?

kswanson33 (changes requested):
`BakeChapterReviewExercises` is being renamed to `MoveExercisesToEOC` to reduce confusion about exercise directions. I've made a [PR](https://github.com/openstax/kitchen/pull/195) for you; if this looks good, feel free to merge it into the current PR to take care of the renaming on this branch.

lib/kitchen/directions/bake_chapter_review_exercises/v1.rb
kswanson33: unless I'm missing something...

wilkus27: Thank you for catching this :relaxed: I will try to add all instances to specs.

wilkus27:
@LostMojo @kswanson33 I added specs for both versions. PR is ready for review. I'm new to creating rspecs, so I'm not sure if this is how it suppose to look :grin: Let me know if I did something wrong.

kswanson33 (approved):
This looks great (& should fix a problem I was having with my book)! Thank you!

LostMojo (approved):
Awesome! This looks great, and I can't wait to use it!

Issue #176, opened 2021-04-12
Refactor some_element.document.pantry to some_element.pantry


jpslav:
There are 20-30 places in the code where we access a pantry or clipboard through an element's document:

```ruby
figure.document.pantry(name: :link_text)...
```

Ditto for `clipboard`

Add 

```ruby
def_delegators :document, :pantry, :clipboard
```

to `ElementBase` and search and replace all `.document.pantry` to `.pantry` and `.document.clipboard` to `.clipboard`.

PR #196, opened 2021-04-26
refactor access to pantry

brittany-johnson:
https://github.com/openstax/kitchen/issues/176

### DIFF
--- a/lib/kitchen/element_base.rb
+++ b/lib/kitchen/element_base.rb
@@ -391,7 +391,7 @@ module Kitchen
       block_error_if(block_given?)
 
       node.remove
-      clipboard(to).add(self) if to.present?
+      get_clipboard(to).add(self) if to.present?
       self
     end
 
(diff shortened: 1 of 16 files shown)


lib/kitchen/element_base.rb
```diff
@@ -92,6 +92,14 @@ class ElementBase
     #   @return [Selectors::Base]
     def_delegators :config, :selectors
 
+    # @!method pantry
+    #   Access the pantry for this element's document
+    #   @return [Pantry]
+    # @!method :clipboard
+    #   Access the clipboard for this element's document
+    #   @return [Clipboard]
+    def_delegators :document, :pantry, :clipboard
```
brittany-johnson, line 101: @jpslav I added a `:clipboard` but I couldn't actually update any calls to `document.clipboard` because on line 682 `clipboard` is a protected method.
Should I remove `:clipboard` from line 101 then or do you have another suggestion?

jpslav (approved):

PR #197, opened 2021-04-28
Add terms from composite pages to index

kswanson33:
Currently, easybake handles terms that appear in composite pages by acting as if they originated from the final page in the chapter. This feature:

1. ensures terms in composite pages are not overlooked in kitchen
2. updates the link text for the key term callback in the index to give the chapter number & eoc section name, instead of the final page of the chapter

image
^ easybake version

image
^ kitchen version

### DIFF
--- a/lib/kitchen/directions/bake_index/v1.rb
+++ b/lib/kitchen/directions/bake_index/v1.rb
@@ -119,26 +119,38 @@ module Kitchen::Directions::BakeIndex
       @index = Index.new
 
       book.pages.terms.each do |term_element|
-        # Markup the term
         page = term_element.ancestor(:page)
         term_element.id = "auto_#{page.id}_term#{term_element.count_in(:book)}"
+        page_title = page.title.text
+        add_term_to_index(term_element, page_title)
+      end
 
-        group_by = term_element.text.strip[0]
-        group_by = I18n.t(:eob_index_symbols_group) unless group_by.match?(/\w/)
-        term_element['group-by'] = group_by
-
-        # Add it to our index object
-        @index.add_term(
-          Term.new(
-            text: term_element.text,
-            id: term_element.id,
-            group_by: group_by,
-            page_title: page.title.text.gsub(/\n/, '')
-          )
-        )
+      book.chapters.composite_pages.terms.each do |term_element|
+        page = term_element.ancestor(:composite_page)
+        chapter = term_element.ancestor(:chapter)
+        term_element.id = "auto_composite_page_term#{term_element.count_in(:book)}"
+        chapter_number = chapter.count_in(:book)
+        page_title = "#{chapter_number} #{page.title.text.strip}".strip
+        add_term_to_index(term_element, page_title)
       end
 
       book.first('body').append(child: render(file: 'v1.xhtml.erb'))
+    end
+
+    def add_term_to_index(term_element, page_title)
+      group_by = term_element.text.strip[0]
+      group_by = I18n.t(:eob_index_symbols_group) unless group_by.match?(/\w/)
+      term_element['group-by'] = group_by
+
+      # Add it to our index object
+      @index.add_term(
+        Term.new(
+          text: term_element.text,
+          id: term_element.id,
+          group_by: group_by,
+          page_title: page_title.gsub(/\n/, '')
+        )
+      )
     end
 
   end
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+* Add terms from composite pages to index (minor)
 * Access `.pantry` and `.clipboard` through an element instead of just its document (minor)
 
 ## [3.2.0] - 2021-04-19
--- a/lib/kitchen/composite_page_element.rb
+++ b/lib/kitchen/composite_page_element.rb
@@ -26,7 +26,7 @@ module Kitchen
       # Get the title in the immediate children, not the one in the metadata.  Could use
       # CSS of ":not([data-type='metadata']) > [data-type='document-title'], [data-type='document-title']"
       # but xpath is shorter
-      first!("./*[@data-type = 'document-title']")
+      first!("./*[@data-type = 'document-title' or @data-type = 'title']")
     end
 
     # Returns true if this class represents the element for the given node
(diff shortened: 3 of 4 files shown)


lib/kitchen/element_enumerator_base.rb
```diff
@@ -101,6 +101,20 @@ def non_introduction_pages(only: nil, except: nil)
...
+    #   iterated over.
+    #
+    def pages_and_composite_pages(only: nil, except: nil)
+      block_error_if(block_given?)
+      chain_to(PageElementEnumerator,
+               css_or_xpath: "div[data-type='page'], div[data-type='composite-page']",
+               only: only,
+               except: except)
+    end
+
```
kswanson33, line 117: Is there a way to dynamically cast elements as either PageElements or CompositePageElements?

jpslav: Yes, there is a `TypeCastingElementEnumerator`.  Also `ElementBase#search_with`, which maybe is unused at the moment.  I'd look at these, might need to add `except` and `only` to `search_with`

jpslav (changes requested):

lib/kitchen/directions/bake_index/v1.rb
```diff
@@ -118,10 +118,21 @@ def bake(book:)
...
+      chapters = book.chapters
+      book.pages_and_composite_pages.terms.each do |term_element|
         # Markup the term
         page = term_element.ancestor(:page)
-        term_element.id = "auto_#{page.id}_term#{term_element.count_in(:book)}"
+        term_element.id = ''
+        page_title = ''
+        if page[:'data-type'] == 'composite-page'
+          term_element.id = "auto_composite_page_term#{term_element.count_in(:book)}"
+          chapter_number = chapters.search("[id='#{term_element.id}']")&.first&.ancestor(:chapter)&.count_in(:book)
```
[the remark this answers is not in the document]
kswanson33: Terms are not always in a chapter; they seem to be able to come from the Preface and Introduction as well. But if the term is in a composite page, then it is in a chapter. So I didn't use `book.chapters` here because that would exclude those pages. Believe me, I wouldn't have gotten the chapter number this way otherwise 🙃 

Since `chapters` is an `ElementEnumerator`, not an `Element`, `first` actually calls the base enumerator `first` method. I don't think that should be overwritten, so I think `search(...).first` has to be used here (again, 🙃)

lib/kitchen/directions/bake_index/v1.rb
```diff
@@ -118,10 +118,21 @@ def bake(book:)
...
+          title_text = page.first("[data-type='document-title'], [data-type='title']").text.strip
```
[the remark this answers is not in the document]
kswanson33: if the page is cast to the appropriate type (page vs composite page) then yes ... I will look into that

lib/kitchen/directions/bake_index/v1.rb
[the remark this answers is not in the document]
jpslav: ah you're right on `first`

lib/kitchen/directions/bake_index/v1.rb
```diff
@@ -118,10 +118,25 @@ def bake(book:)
...
+      chapters = book.chapters
+      book.pages_and_composite_pages.terms.each do |term_element|
         # Markup the term
-        page = term_element.ancestor(:page)
-        term_element.id = "auto_#{page.id}_term#{term_element.count_in(:book)}"
+        begin
+          page = term_element.ancestor(:page)
+        rescue RuntimeError
+          page = term_element.ancestor(:composite_page)
+        end
```
jpslav, line 128: [Use exceptions for exceptional behavior only](https://bit.ly/3eBz0Dm), not for normal flow behavior.

I bet you look at this code and think it looks kind of gnarly.  The world is your oyster, so can you envision other approaches that would look better and make you feel better about the code, then see if you can implement them?  I can think of two ways.  **Take a look at them after** you ponder this for yourself a bit: https://gist.github.com/jpslav/7a53c1a4c0a17d2a735f6fd87abb7abc

lib/kitchen/directions/bake_index/v1.rb
```diff
@@ -118,10 +118,25 @@ def bake(book:)
...
+        page_title = ''
+        if page.instance_of?(Kitchen::CompositePageElement)
+          term_element.id = "auto_composite_page_term#{term_element.count_in(:book)}"
+          chapter_number = chapters.search("[id='#{term_element.id}']")&.first&.ancestor(:chapter)&.count_in(:book)
+          title_text = page.title.text.strip
+          page_title = "#{chapter_number} #{title_text}".strip
+        else
+          term_element.id = "auto_#{page.id}_term#{term_element.count_in(:book)}"
+          page_title = page.title.text
+        end
```
jpslav, line 139: Just realized this other way to make this cleaner -- two loops -- would mean the term counts in book aren't unique, but the overall ID would still be

```ruby
def bake(book:)
  ...
  book.pages.terms.each do |term_element|
    page = term_element.ancestor(:page)
    term_element.id = "auto_#{page.id}_term#{term_element.count_in(:book)}"
    page_title = page.title.text
    add_to_index(term_element, page_title)
  end
  
  book.chapters.composite_pages.terms.each do |term_element|
    page = term_element.ancestor(:composite_page)
    chapter = term_element.ancestor(:chapter)
    term_element.id = "auto_composite_page_term#{term_element.count_in(:book)}"
    title_text = page.title.text.strip
    page_title = "#{chapter.count(in: :book)} #{title_text}".strip
    add_to_index(term_element, page_title)
  end
  ...
end

def add_to_index(term_element, page_title)
  group_by = term_element.text.strip[0]
  group_by = I18n.t(:eob_index_symbols_group) unless group_by.match?(/\w/)
  term_element['group-by'] = group_by

  # Add it to our index object
  @index.add_term(
    Term.new(
      text: term_element.text,
      id: term_element.id,
      group_by: group_by,
      page_title: page_title.gsub(/\n/, '')
    )
  )
end
```

jpslav (changes requested):

lib/kitchen/directions/bake_index/v1.rb
kswanson33: wow thank you mr mentor

kswanson33: To show you that I did my homework, my ideas for a different way of doing this were along the lines of "well I could safely inquire about what ancestors are available on an element with `element.ancestors.keys` or use the `has_ancestor?` method, to avoid triggering the exception." I tried to think of a second way for kicks, but I was never any good at not peeking at the upside down text on those brainteaser mysteries as a kid. I think my solution would be more of a bandaid in comparison to your proposed solutions, with all their good refactoring in base Kitchen.

I see you've assigned the second of the two approaches to me in a separate issue. I also see your comment below outlining a two-loop solution. Does it work if I implement your comment below for this PR, and do the yield-multiple-elements refactor later, on a different branch?

Thanks so much for working out these options!

jpslav (approved):

PR #198, opened 2021-04-28
Eoc free response

LostMojo:
Moves free response questions to the end of the chapter & creates a container for them

### DIFF
--- a/lib/kitchen/directions/bake_free_response/v1.rb
+++ b/lib/kitchen/directions/bake_free_response/v1.rb
@@ -24,6 +24,7 @@ module Kitchen::Directions::BakeFreeResponse
       end
 
       append_to_element = append_to || chapter
+      @title_tag = append_to ? 'h3' : 'h2'
 
       append_to_element.append(child: render(file: 'free_response.xhtml.erb'))
     end
--- a/lib/kitchen/directions/bake_free_response/v1.rb
+++ b/lib/kitchen/directions/bake_free_response/v1.rb
@@ -4,8 +4,28 @@ module Kitchen::Directions::BakeFreeResponse
   class V1
     renderable
 
-    def bake(chapter:, metadata_source:, klass:, append_to: nil)
-      #bloop
+    def bake(chapter:, metadata_source:, append_to:)
+      @metadata_elements = metadata_source.children_to_keep.copy
+
+      @free_response_questions = []
+      free_response_clipboard = Kitchen::Clipboard.new
+      chapter.non_introduction_pages.each do |page|
+        free_response_questions = page.free_response
+        next if free_response_questions.none?
+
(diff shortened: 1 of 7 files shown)


LostMojo:
The legacy baked free response is a composite page container containing a title, metadata and exercises (that are baked by another process) and it looks like this:

Screen Shot 2021-05-03 at 9 14 50 AM

lib/kitchen/directions/bake_free_response/v1.rb
```diff
@@ -0,0 +1,32 @@
...
+
+        free_response_questions.search('h3').trash
+        title = Kitchen::Directions::EocSectionTitleLinkSnippet.v1(page: page)
+        free_response_questions.each do |free_response_question|
+          free_response_question.prepend(child: title)
+          free_response_question&.cut(to: free_response_clipboard)
+        end
+        @free_response_questions.push(free_response_clipboard.paste)
+        free_response_clipboard.clear
+      end
```
jpslav, line 24: A clipboard is essentially an array, so I'm wondering if we don't need both an array and a clipboard on lines 10 and 11.

```ruby
def bake(chapter:, metadata_source:, append_to:)
  @metadata_elements = metadata_source.children_to_keep.copy

  @free_response_clipboard = Kitchen::Clipboard.new
    free_response_questions = page.free_response
    next if free_response_questions.none?

    free_response_questions.search('h3').trash
    title = Kitchen::Directions::EocSectionTitleLinkSnippet.v1(page: page)
    free_response_questions.each do |free_response_question|
      free_response_question.prepend(child: title)
      free_response_question&.cut(to: @free_response_clipboard)
    end
  end

  append_to_element = append_to || chapter
  @title_tag = append_to ? 'h3' : 'h2'

  append_to_element.append(child: render(file: 'free_response.xhtml.erb'))
end
```

and then do `@free_response_clipboard.paste` in the `.xhtml.erb` file?

lib/kitchen/directions/bake_free_response/v1.rb
```diff
@@ -0,0 +1,32 @@
...
+      free_response_clipboard = Kitchen::Clipboard.new
+      chapter.non_introduction_pages.each do |page|
+        free_response_questions = page.free_response
+        next if free_response_questions.none?
+
+        free_response_questions.search('h3').trash
+        title = Kitchen::Directions::EocSectionTitleLinkSnippet.v1(page: page)
+        free_response_questions.each do |free_response_question|
+          free_response_question.prepend(child: title)
+          free_response_question&.cut(to: free_response_clipboard)
```
jpslav, line 20: Guessing you don't need the `&` here since it isn't used on line 19

jpslav (changes requested):

jpslav (approved):

PR #199, opened 2021-04-29
suppress solutions

wilkus27:
Add option to `bake_numbered_exercise` to be able to suppress solutions.


lib/kitchen/directions/bake_numbered_exercise/v1.rb
```diff
@@ -2,15 +2,19 @@
...
 
       if solution.present?
-        problem_number = "<a class='os-number' href='##{exercise.id}-solution'>#{number}</a>"
-        bake_solution(exercise: exercise, number: number)
+        if !suppress_solution
+          problem_number = "<a class='os-number' href='##{exercise.id}-solution'>#{number}</a>"
+          bake_solution(exercise: exercise, number: number)
+        elsif suppress_solution
+          solution.trash
+        end
```
kswanson33, line 17: what about

```ruby
if solution.present? and !suppress_solution
  # number & bake solution...
elsif suppress_solution
  # trash solution
end
```

kswanson33 (changes requested):
One comment, but otherwise looks good!

lib/kitchen/directions/bake_numbered_exercise/v1.rb
wilkus27: @kswanson33 It doesn't work, because for the `elsif` statement we also need to have solutions, to be able to trash them.
Here is the error:

image

That's why I applied nested `if` statements.
I can change it to something like that:

image

if we want one `if` statement. Let me know what you think :wink:

kswanson33: I like the second option, keeping it to one `if` statement is good! Another possibility is to use ruby's [safe navigation operator](https://mitrev.net/ruby/2015/11/13/the-operator-in-ruby/). This would let you write `solution&.trash`.

wilkus27:
@kswanson33 Thank you for an advice with safe navigation operator :slightly_smiling_face: PR is ready for rereview.

kswanson33 (approved):

lib/kitchen/directions/bake_numbered_exercise/v1.rb
```diff
@@ -2,15 +2,17 @@
...
       solution = exercise.solution
 
       problem_number = "<span class='os-number'>#{number}</span>"
 
-      if solution.present?
+      if solution.present? && !suppress_solution
         problem_number = "<a class='os-number' href='##{exercise.id}-solution'>#{number}</a>"
         bake_solution(exercise: exercise, number: number)
+      elsif suppress_solution
+        solution&.trash
```
jpslav, line 15: To not have to check `suppress_solution` twice, you can:

```ruby
if solution.present? 
  if suppress_solution
    solution.trash
  else
    problem_number = "<a class='os-number' href='##{exercise.id}-solution'>#{number}</a>"
    bake_solution(exercise: exercise, number: number)
  end
end
```

Also since you know that `solution.present?` in this way you don't need the `&` safe navigation operator.  

These two changes make this code reads a little more clearly to me.  Fewer checks.

ajluk (approved):

jpslav (changes requested):

lib/kitchen/directions/bake_numbered_exercise/v1.rb
jpslav: Also I realize this reverses Kendra's comment. :-)

kswanson33: okay this is probably better... my bad natalia

wilkus27: No problem :wink:

wilkus27:
@jpslav Thank you for comments. I added requested changes. PR is ready for review.

jpslav (approved):

jpslav:
Thanks @wilkus27 !

PR #200, opened 2021-04-29
add amgov answer key strategy

brittany-johnson:
related card https://github.com/openstax/cnx-recipes/issues/3506

### DIFF
--- a/lib/kitchen/directions/move_solutions_to_answer_key/strategies/american_government.rb
+++ b/lib/kitchen/directions/move_solutions_to_answer_key/strategies/american_government.rb
@@ -1,0 +1,30 @@
+# frozen_string_literal: true
+
+module Kitchen::Directions::MoveSolutionsToAnswerKey
+  module Strategies
+    class AmericanGovernment
+      def bake(chapter:, append_to:)
+        bake_section(chapter: chapter, append_to: append_to, klass: 'review-questions')
+      end
+
+      protected
+
+      def bake_section(chapter:, append_to:, klass:)
+        section_solutions_set = []
+        chapter.search(".#{klass}").each do |section|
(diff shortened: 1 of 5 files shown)


lib/locales/en.yml
```diff
@@ -23,6 +23,10 @@ en:
   eoc_key_concepts: Key Concepts
   eoc_key_equations: Key Equations
   eoc_suggested_reading: Suggestions for Further Study
+  eoc: {
+    review-questions: Review Questions,
+    critical-thinking: Critical Thinking Questions
+  }
```
jpslav, line 29: Why are these added if they aren't used in this PR?

brittany-johnson: @jpslav In my recipe, I'm using `MoveExercisesToEOC.v1` and passing these values to it. I included it in this PR because both of these sections have solutions in my answer key.

lib/kitchen/directions/move_solutions_to_answer_key/strategies/american_government.rb
```diff
@@ -0,0 +1,30 @@
...
+        append_solution_area(section_solutions_set, append_to)
+      end
+
+      def append_solution_area(solutions, append_to)
+        solutions.each do |solution|
+          append_to.add_child(solution.raw)
+        end
+      end
+    end
+  end
```
jpslav, line 29: Actually can't these two methods be rewritten as one method:

```ruby
def bake_section(chapter:, append_to:, klass:)
  chapter.search(".#{klass} [data-type='solution']").each do |solution|
    append_to.append(child: solution.cut.to_s)
  end
end
```

jpslav (changes requested):

jpslav (approved):

PR #204, opened 2021-05-05
Change summary selector in Bake Chapter Summary for sociology recipe

ajluk:
In Sociology 2e summary selector is changed (overridden in recipe) to `.section-summary` to spot it but similar change is also needed for `container` class and `data-uuid-key`: 
image
The title name is just replaced to Section Summary in locales: 
image
Related card:  https://github.com/openstax/cnx-recipes/issues/3504


jpslav:
Do we have any idea if this change will be used in other books beyond Sociology 2e?

If there are other books with different summary classes and data UUID keys, I'd rather have the interface look like this:

```ruby
def self.v1(chapter:, metadata_source:, 
            summary_class: 'os-section-summary-container', summary_data_uuid_key: '.section-summary')
```

so that it is more explicit how things are being overridden -- from the recipe for Sociology 2e you could see that something different was going on, rather than seeing a flag being passed in and not knowing what it was doing.  Also, if a flag is used here and there are differences in other books that don't fit the pattern used in the flag approach, the interface will get complicated.

If Sociology 2e is the only book that has this difference, given that it is an old book I might opt for an approach that doesn't change this direction code but rather does a post processing cleanup step in the recipe.  Either like this:

```ruby
Bake ChapterSummary.v1(chapter: chapter, metadata_source: metadata_source)

# Sociology 2e uses different tags from all other books
book.search('.os-summary-container').each do |element| 
  element.remove_class('os-summary-container')
         .add_class('os-section-summary-container')
         .set('data-uuid-key', '.section-summary')
end
```

Another alternative which I haven't thought through all the way is to add in a configuration per direction kind of like how i18n works (and was easy to fix here)

```ruby
module BakeChapterSummary
  config do
    summary_class { 'os-summary-container' }
    data_uuid_key { '.summary' }
  end

  def self.v1(...)
    ...
    chapter.append(child:
      <div class="os-eoc #{config.summary_class}" data-type="composite-page" data-uuid-key=".#{config.data_uuid_key}">
    ...
  end

where in your recipe you could have a `config.yml` file that had these overrides

```yaml
BakeChapterSummary:
  summary_class: os-summary-container
  data_uuid_key: .summary
```

Lots of options here.  Curious what @KarinaMendez2 @KarinaMendez2 @LostMojo @brittany-johnson think

ajluk:
@jpslav There are few books that utilize also `section-summary` class for Summary sections (f.ex. College Physics, Anthropology, Business Ethics, World History, Entrepreneurship, Principles of Management ). But there are no other different classes used for Summary in any other books as far as I know. There is either `summary` or `section-summary` class used for Summary in all of the books.

brittany-johnson:
@ajluk I would have gone with the first approach too - we can pass the two values to override when we need it in the recipe.

ajluk:
@jpslav @brittany-johnson the PR is ready for review.

ajluk:
I have updated this PR and adjust my changes to recent EoC changes merged to main branch made by MaryJo (#207, #213). The PR is ready for review.

jpslav (approved):

PR #205, opened 2021-05-05
Create indexes of different types

wilkus27:


wilkus27:
In Polish Psychology there are three types of indexes. Regular one, with names and with foreign terms. In `cnx-recipes` we used namespaces to select proper terms:

image

Namespaces were declared in this way:

image

I'm not sure how can I define them in Ruby. I'm getting such error now:

image

This is how I'm using namespace in code:

image

wilkus27:
Here is how terms look in `assembled.xtml`:

image

image

image

And here is `baked.xhtml`:

image

I will create specs for that, but since it might take a moment I'm adding screenshots for now for better understanding of the problem.

wilkus27:
This PR is outdated, so I'm closing it. Work for indexes is done here: https://github.com/openstax/kitchen/pull/300

PR #207, opened 2021-05-07
Single Template for EOC sections

LostMojo:
This change will alter most of the eoc chapter content with similar layouts to pull from one template. (This also solves for h2/h3 differences)


lib/kitchen/directions/move_exercises_to_eoc/v1.rb
```diff
@@ -33,10 +34,11 @@ def bake(chapter:, metadata_source:, klass:, append_to: nil)
       @content = exercise_clipboard.paste
 
       append_to_element = append_to || chapter
-      @tag = append_to ? 'h3' : 'h2'
+      @in_composite_chapter = append_to.present?
       @metadata_title = append_to ? I18n.t(:eoc_exercises_title) : @title
```
kswanson33, line 38: can this line be removed since this is sorted out in the template

kswanson33 (changes requested):
🔥🔥 Thank you for doing this!!

Do you think the changes here need tests? I don't think a full instance of each EOC section is necessary, but it might be good to spec just the titles for both the append_to/no append_to cases.

lib/kitchen/directions/move_exercises_to_eoc/v1.rb
LostMojo: Fixed that! And yeah, I think it'll be worth it to have spec for the composite/no composite chapters

lib/kitchen/directions/bake_chapter_glossary.rb
```diff
@@ -5,35 +5,40 @@ module Directions
...
-          HTML
-        )
+          definitions = chapter.glossaries.search('dl').cut
+          definitions.sort_by! do |definition|
+            [definition.first('dt').text.downcase, definition.first('dd').text.downcase]
+          end
+
+          chapter.glossaries.trash
+
+          return if definitions.none?
```
jpslav, line 32: I'd move this right after line 25.  If there are no definitions, we don't need to sort them.  Also move `chapter.glossaries.trash` above setting `definitions` b/c it isn't related to these lines of code, seems out of place wandering here in the midst of them.

lib/kitchen/directions/move_exercises_to_eoc/v1.rb
```diff
@@ -33,10 +34,10 @@ def bake(chapter:, metadata_source:, klass:, append_to: nil)
       @content = exercise_clipboard.paste
 
       append_to_element = append_to || chapter
-      @tag = append_to ? 'h3' : 'h2'
-      @metadata_title = append_to ? I18n.t(:eoc_exercises_title) : @title
+      @in_composite_chapter = append_to.present?
```
jpslav, line 37: See slack thread about this, in a bunch of these files

jpslav (changes requested):

lib/kitchen/directions/bake_chapter_key_concepts/v1.rb
```diff
@@ -16,16 +18,18 @@ def bake(chapter:, metadata_source:, append_to:)
...
         end
-        @key_concepts.push(key_concepts_clipboard.paste)
-        key_concepts_clipboard.clear
       end
 
+      @content = "<div class=\"os-key-concepts\"> #{key_concepts_clipboard.paste} </div>"
+
       append_to_element = append_to || chapter
-      @title_tag = append_to ? 'h3' : 'h2'
+      @in_composite_chapter = append_to_element[:'data-type'] == 'composite-chapter'
```
jpslav, line 29: so that we keep the `data-type` knowledge all in one place, can you add an `is?(type)` method to `ElementBase`?  Has a few steps

```ruby
class ElementBase
  def self.descendant(type)
    @types_to_descendants ||=
      descendants.each_with_object({}) do |descendant, hash|
        hash[descendant.short_type] = descendant
      end

    @types_to_descendants[type]
  end

  def self.descendant!(type)
    descendant(type) || raise("Unknown ElementBase descendant type '#{type}'")
  end

  def is?(type)
    self.class.descendant!(type).is_the_element_class_for?(raw)
  end
end
```

`is_the_element_class_for?` already exists in each element class.  Then add a class method `short_type` to each element class, e.g. changing 

```ruby
class CompositeChapterElement < ElementBase
  def initialize(node:, document: nil)
      super(node: node,
            document: document,
            enumerator_class: ElementEnumerator,
            short_type: :composite_chapter)
    end
```

to

```ruby
class CompositeChapterElement < ElementBase
  def self.short_type
    :composite_chapter
  end

  def initialize(node:, document: nil)
      super(node: node,
            document: document,
            enumerator_class: ElementEnumerator,
            short_type: self.class.short_type)
    end
```

And then this line becomes

```ruby
@in_composite_chapter = append_to_element.is?(:composite_chapter)
```

jpslav (approved):

PR #215, opened 2021-05-13
Rework copy, cut, paste methods

brittany-johnson:


lib/kitchen/document.rb
```diff
@@ -44,7 +44,7 @@ def initialize(nokogiri_document:, config: nil)
       @nokogiri_document = nokogiri_document
       @location = nil
       @config = config || Config.new
-      @next_paste_count_for_id = {}
+      @id_cp_count = Hash.new { |hash, key| hash[key] = {} }
```
jpslav, line 47: If you change this to 

```ruby
@id_cp_count = Hash.new { |hash, key| hash[key] = {count: 0, last_paste: false} }
```

it'll give you the initial values you want, lets you skip some lines below

lib/kitchen/document.rb
```diff
@@ -165,7 +165,32 @@ def create_element_from_string(string)
     def record_id_copied(id)
       return if id.blank?
 
-      @next_paste_count_for_id[id] ||= 1
+      @id_cp_count[id][:count]&.positive? ? @id_cp_count[id][:count] += 1 : @id_cp_count[id][:count] = 1
```
jpslav, line 168: With the above change, and the fact that line 179 below prevents negative counts from ever happening, this reduces to 

```ruby
@id_cp_count[id][:count] += 1
```

lib/kitchen/document.rb
```diff
@@ -165,7 +165,32 @@ def create_element_from_string(string)
...
+    end
+
+    # Keeps track that an element with the given ID has been cut.
+    #
+    # @param id [String]
+    #
+    def record_id_cut(id)
+      return if id.blank?
+
+      @id_cp_count[id][:count]&.positive? ? @id_cp_count[id][:count] -= 1 : @id_cp_count[id][:count]
```
jpslav, line 179: Note in this line that the "else" condition after the `:` isn't doing anything.  It isn't assigning anything, nor is the value it returns being assigned anywhere.  So you can simplify this to

```ruby
@id_cp_count[id][:count] -= 1 if @id_cp_count[id][:count].positive?
```

lib/kitchen/document.rb
```diff
@@ -165,7 +165,32 @@ def create_element_from_string(string)
...
+      @id_cp_count[id][:last_paste] = false
+    end
+
+    # Keeps track that an element with the given ID has been pasted.
+    #
+    # @param id [String]
+    #
+    def record_id_paste(id)
```
jpslav, line 187: to match the other methods, this should end `_pasted` (past tense).  Also `:last_paste` should be `:last_pasted`.

lib/kitchen/document.rb
```diff
@@ -165,7 +165,32 @@ def create_element_from_string(string)
...
+      return if id.blank?
+
+      @id_cp_count[id][:count] ||= 0
```
jpslav, line 190: With the change on line 47 above, this line can be deleted.

lib/kitchen/document.rb
```diff
@@ -165,7 +165,32 @@ def create_element_from_string(string)
...
+      @id_cp_count[id][:last_paste] ? @id_cp_count[id][:count] += 1 : @id_cp_count[id][:count]
```
jpslav, line 191: Like above the part after the `:` isn't doing anything, so change to

```ruby
@id_cp_count[id][:count] += 1 if @id_cp_count[id][:last_paste]
```

lib/kitchen/document.rb
```diff
@@ -177,8 +202,7 @@ def modified_id_to_paste(original_id)
       return nil if original_id.nil?
       return '' if original_id.blank?
 
-      count = next_count_for_pasted_id(original_id)
-
+      count = @id_cp_count[original_id][:count] ||= 0
```
jpslav, line 205: With the change on line 47 above, the `||= 0` can be removed.

lib/kitchen/element_base.rb
```diff
@@ -429,6 +429,12 @@ def search_with(*enumerator_classes)
     def cut(to: nil)
       block_error_if(block_given?)
 
+      temp = clone
```
jpslav, line 432: In `paste`, a `clone` is needed because we are changing the IDs on line 474.  Here we aren't changing this element so we don't need (or want) a clone.  You can simply do 

```ruby
raw.traverse do |node|        
  next if node.text? || node.document?        
  document.record_id_cut(node[:id])      
end
```

lib/kitchen/element_base.rb
```diff
@@ -460,11 +466,11 @@ def copy(to: nil)
     def paste
       # See `clone` method for a note about namespaces
       block_error_if(block_given?)
-
       temp_copy = clone
       temp_copy.raw.traverse do |node|
         next if node.text? || node.document?
 
+        document.record_id_paste(node[:id])
         node[:id] = document.modified_id_to_paste(node[:id]) unless node[:id].blank?
```
jpslav, line 474: I know that `record_id_paste` is doing a check for blank IDs, but it seems a little weird here to a some other developer reading this to see a blank check on line 474 but not on 473.  Can we wrap both lines in one check?

```ruby
if node[:id].present?
  document.record_id_paste(node[:id])        
  node[:id] = document.modified_id_to_paste(node[:id])
end
```

lib/kitchen/document.rb
```diff
@@ -177,8 +202,7 @@ def modified_id_to_paste(original_id)
       return nil if original_id.nil?
```
jpslav, line 202: Ok, so feel free to hate me for saying this, but after you make the other changes... here's a refactoring that I'd like...

Do you see how all of this ID tracking code has nothing to do with the `Document` class?  It literally does not use any state of the `Document` class aside from `@id_cp_count`, and these are the only `Document` methods that use that hash.  

So to me, this yells "This code belongs in a different class!"  And yes, I realize that I put us down this path, but it is never too late for me to recognize my own mistakes :-)

So make a new class:

```ruby
class IdTracker
  def initialize(copy_suffix: '_copy_'); ... end
  def record_copied(id); ... end
  def record_cut(id); ... end
  def record_pasted(id); ... end
  def modified_id_to_paste(id); ... end
end
```

Remember to require it in `openstax_kitchen.rb`.

Aside: note how in this PR `record_id_pasted` and `modified_id_to_paste` are called together right next to each other -- maybe it should just be one method in this new class called `value_to_paste(id)` or similar.  Or `record_pasted` just moves inside `modified_id_to_paste`

Write a unit test for this class.  It should be easier than normal b/c you don't have to construct any XML.  You can just say `record_copied('foo')`, `expect(modified_id_to_paste('foo').to eq 'foo_copy_1'`, etc

Then in `Document`

```ruby
class Document
  attr_reader :id_tracker
  def initialize(...)
    @id_tracker = IdTracker.new
    # can get rid of the `@id_copy_suffix` b/c now it is inside `IdTracker`
  end
```

and in `ElementBase`

```ruby
def_delegators :document, :id_tracker
```

which will let you call `id_tracker.record_copied(id)`, etc.

jpslav (changes requested):

lib/kitchen/id_tracker.rb
```diff
@@ -0,0 +1,67 @@
+# frozen_string_literal: true
+
+module Kitchen
+  # A class to track and modify duplicate IDs in a document
+  #
+  class IdTracker
+
+    def initialize
+      @id_cp_count = Hash.new { |hash, key| hash[key] = { count: 0, last_pasted: false } }
```
jpslav, line 9: this variable is more than a "copy count" now - suggest renaming to something like `@id_data`

jpslav (changes requested):

spec/directions/move_solutions_to_answer_key/calculus_strategy_spec.rb
```diff
@@ -203,7 +203,7 @@
               <h3 data-type="title">
                 <span class="os-title-label">Review Exercises</span>
               </h3>
-              <div data-type="solution" id="auto_m68764_fs-idm128797504_copy_1">
+              <div data-type="solution" id="auto_m68764_fs-idm128798615">
```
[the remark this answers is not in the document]
brittany-johnson: @jpslav In `book1` there were two elements that had the same ID (Line 66 and 89) which shouldn't happen in real content

jpslav (approved):

PR #221, opened 2021-05-17
Add Eoc Reference 

LostMojo:
Bakes an end of chapter Reference section using the EOC title template

### DIFF
--- a/lib/kitchen/directions/bake_chapter_references/v1.rb
+++ b/lib/kitchen/directions/bake_chapter_references/v1.rb
@@ -24,8 +24,7 @@ module Kitchen::Directions::BakeChapterReferences
         end
       end
 
-      references = chapter.pages.references.cut
-      @content = references.paste
+      @content = chapter.pages.references.cut.paste
 
       @in_composite_chapter = false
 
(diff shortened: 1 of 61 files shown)


lib/kitchen/directions/bake_chapter_references/v1.rb
```diff
@@ -0,0 +1,36 @@
...
+
+        references.search('h3').trash
+        title = Kitchen::Directions::EocSectionTitleLinkSnippet.v1(page: page)
+
+        references.each do |reference|
+          reference.prepend(child: title)
+        end
+      end
+
+      references = chapter.pages.references.cut
```
KarinaMendez2, line 27: If you use this in the template, I think it could be just one like, put an “@“ to references and paste it in the template

LostMojo: Done!

LostMojo: Since this uses the common eoc template, I have to go by it's rules, and it takes a string, not a clipboard

KarinaMendez2: oh right, approved!

KarinaMendez2 (approved):

PR #223, opened 2021-05-18
Enumerator selectors

jpslav:
The selectors object within a document's config object is intended to be a common place to store all of the main CSS or XPath selectors, e.g. those that define chapters, notes, pages, etc.

A nice thing about such a common place is that it gives us easy access to override the selectors to account for book-by-book differences.  E.g. some books identify references with a "reference" class (singular), while others use a "references" class (plural).  

Really using the selectors means we need to be able to use them from enumerators.  When enumerators are used and chained, we don't immediately have access to the config object which contains the selectors -- we don't get that access until we are actually running a search on an element and can get to the config via the element's document.  

When we define an enumerator, we define its `default_css_or_xpath`, e.g. `[data-type="page"]`.  So that we can define this value in the selectors and use them down in the guts of enumeration, this PR changes the `default_css_or_xpath` to optionally be a `Proc`, aka a little function that can accept the `config` object and call `config.selectors.something` when the `config` object is available.

You can pass these Procs into the `ElementEnumeratorFactory` directly, e.g. 

```ruby
ElementEnumeratorFactory.new(
  default_css_or_xpath: ->(config) { config.selectors.reference},
  sub_element_class: PageElement,
  enumerator_class: self
)
```

or you can use the new `Selector` class to give you such a `Proc` in a neater form:

```ruby
default_css_or_xpath: Selector.named(:reference)
```

A bonus of this new `Selector` object is that we can also use it to implement the `is_the_element_class_for?` method without duplicating selector values, e.g.

```ruby
def self.is_the_element_class_for?(node, config:)
  Selector.named(:reference).matches?(node, config: config)
end
```

A lot of the files in this PR changed only to account for the new `config:` argument to `is_the_element_class_for?`.

### DIFF
--- a/lib/kitchen/chapter_element.rb
+++ b/lib/kitchen/chapter_element.rb
@@ -68,14 +68,5 @@ module Kitchen
       search('[data-type="abstract"]')
     end
 
-    # Returns true if this class represents the element for the given node
-    #
-    # @param node [Nokogiri::XML::Node] the underlying node
-    # @return [Boolean]
-    #
-    def self.is_the_element_class_for?(node, config:)
-      Selector.named(:chapter).matches?(node, config: config)
-    end
-
   end
 end
--- a/lib/kitchen/chapter_element.rb
+++ b/lib/kitchen/chapter_element.rb
@@ -73,8 +73,8 @@ module Kitchen
     # @param node [Nokogiri::XML::Node] the underlying node
     # @return [Boolean]
     #
-    def self.is_the_element_class_for?(node, **)
-      node['data-type'] == 'chapter'
+    def self.is_the_element_class_for?(node, config:)
+      Selector.named(:chapter).matches?(node, config: config)
     end
 
   end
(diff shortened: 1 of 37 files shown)


lib/kitchen/chapter_element.rb
```diff
@@ -73,7 +73,7 @@ def abstracts
     # @param node [Nokogiri::XML::Node] the underlying node
     # @return [Boolean]
     #
-    def self.is_the_element_class_for?(node)
+    def self.is_the_element_class_for?(node, **)
```
jpslav, line 76: all of the changes like this are because this method is now passed a `config:` argument that is currently not used (will be used when we add the selector information to the `Selectors` classes).

lib/kitchen/element_enumerator_factory.rb
```diff
@@ -64,6 +65,9 @@ def or_with(other_factory)
     protected
 
     def build_within_element(element, search_query:)
+      search_query.apply_default_css_or_xpath_and_normalize(default_css_or_xpath,
+                                                            config: element.config)
```
jpslav, line 69: this call got moved here so that we can pull the `config` from `element`

LostMojo (commented):
Thank you for doing this! Did you update the changelog?

PR #230, opened 2021-05-21
Precalculus strategy for baking solutions & BakeNumberedNotes v2

kswanson33:


KarinaMendez2 (changes requested):

lib/kitchen/directions/bake_notes/bake_numbered_notes/main.rb
```diff
@@ -0,0 +1,29 @@
...
+      def self.v2(book:, classes:)
+        V2.new.bake(book: book, classes: classes)
+      end
+
+      # Used by V1, V2
+      def self.bake_note_exercise(note:, exercise:, divider: ' ')
+        exercise.add_class('unnumbered')
+        # bake problem
+        exercise.problem.wrap_children('div', class: 'os-problem-container')
+        exercise.problem.first('strong')&.trash
```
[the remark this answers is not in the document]
kswanson33: I believe so; the content change hasn't happened yet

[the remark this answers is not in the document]
KarinaMendez2: then we should figure out the selector thing since this breaks calculus vol2 I think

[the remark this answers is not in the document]
kswanson33: This is just a refactor. I believe the code that breaks calc is already on main, and this PR just moves it to a different place (to allow v1 and v2 of this direction to both access it). To allow work on precalc to progress, I'd advocate for merging this and fixing line 19 in a different PR. What do you think?

lib/kitchen/directions/move_solutions_to_answer_key/strategies/precalculus.rb
```diff
@@ -0,0 +1,86 @@
...
+        classes = %w[review-exercises practice-test]
+        classes.each do |klass|
+          bake_section(chapter: chapter, append_to: append_to, klass: klass)
+        end
+      end
+
+      protected
+
+      def bake_section(chapter:, append_to:, klass:, number: nil)
+        section_solutions_set = []
```
KarinaMendez2, line 26: how about a clipboard?

lib/kitchen/directions/move_solutions_to_answer_key/strategies/precalculus.rb
```diff
@@ -0,0 +1,86 @@
...
+        chapter.search(".#{klass}").each do |section|
+          section.search('[data-type="solution"]').each do |solution|
+            section_solutions_set.push(solution.cut)
```
KarinaMendez2, line 29: so that you can do here `solution.cut(to: section_solutions_set)`

lib/kitchen/directions/move_solutions_to_answer_key/strategies/precalculus.rb
```diff
@@ -0,0 +1,86 @@
...
+        title = <<~HTML
+          <h3 data-type="title">
+            <span class="os-title-label">#{I18n.t(:"eoc.#{klass}", number: number)}</span>
+          </h3>
+        HTML
+
+        append_solution_area(title: title, solutions: section_solutions_set, append_to: append_to)
+      end
+
+      def bake_try_note(chapter:, append_to:)
```
KarinaMendez2, line 44: let's move this into another direction, seams cleaner than to have it in here

KarinaMendez2 (changes requested):

lib/kitchen/directions/move_solutions_to_answer_key/strategies/precalculus.rb
kswanson33: ah, I remember why I had it like this -- I was getting the stray `_copy_XXX` suffix, which should hopefully have been resolved by britt's PR. a clipboard should work!

lib/kitchen/directions/move_solutions_to_answer_key/strategies/precalculus.rb
kswanson33: discussed on call -- I'll rename this for clarity

KarinaMendez2 (approved):

PR #235, opened 2021-05-24
be more specific about deleting numbered note title

kswanson33:


kswanson33:
This is becoming a content change instead

PR #238, opened 2021-05-24
Suggested fixes for Twitter CLDR usage

jpslav:
Note this does not fix all the specs, but it does fix the polish diacritics.

### DIFF
--- a/lib/kitchen/directions/bake_index/v2.rb
+++ b/lib/kitchen/directions/bake_index/v2.rb
@@ -29,13 +29,6 @@ module Kitchen::Directions::BakeIndex
       def initialize(term_text:)
         @term_text = term_text
         @terms = []
-
-        # Sort by transliterated version first to support accent marks,
-        # then by the raw text to support the same text with different capitalization
-        # @sortable = [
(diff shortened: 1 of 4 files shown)


lib/kitchen/directions/bake_index/v2.rb
```diff
@@ -62,8 +49,7 @@ class IndexSection
       def initialize(name:)
         @force_first = name == I18n.t(:eob_index_symbols_group)
         @name = name
-        collator = TwitterCldr::Collation::Collator.new(:pl)
-        @items = SortedSet.new { |a, b| collator.compare(a, b) }
+        @items = SortedSet.new
```
jpslav, line 52: I'm not sure that `SortedSet` takes a block in its constructor, that may have been the root of the ArgumentError.

wilkus27: That was mine prediction, that I'm trying to compare elements in a wrong place :thinking:  I couldn't figure out where I should do this step.

Thank you for helping me with this! :relaxed:

wilkus27 (approved):

PR #243, opened 2021-05-25
way to simplify suppressing even solutions

jpslav:


lib/kitchen/directions/bake_numbered_exercise/main.rb
```diff
@@ -3,18 +3,14 @@
 module Kitchen
   module Directions
     module BakeNumberedExercise
-      def self.v1(exercise:, number:, suppress_solution: false, note_suppressed_solutions: false)
-        V1.new.bake(exercise: exercise, number: number, suppress_solution: suppress_solution,
+      def self.v1(exercise:, number:, suppress_solution_if: false, note_suppressed_solutions: false)
```
ajluk, line 6: @jpslav I have added a small fix above because of this error I've got while baking the recipe with changes from your PR: 
image

spec/directions/bake_numbered_exercise/main_spec.rb
```diff
@@ -5,19 +5,13 @@
 RSpec.describe Kitchen::Directions::BakeNumberedExercise do
   it 'calls v1' do
     expect_any_instance_of(Kitchen::Directions::BakeNumberedExercise::V1).to receive(:bake)
-      .with(exercise: 'exercise1', number: '1.1', suppress_solution: false, note_suppressed_solutions: false)
-    described_class.v1(exercise: 'exercise1', number: '1.1', suppress_solution: false, note_suppressed_solutions: false)
+      .with(exercise: 'exercise1', number: '1.1', suppress_solution_if: false, note_suppressed_solutions: false)
```
ajluk, line 8: Also made the same change as I was written in the above comment in the spec file.

lib/kitchen/directions/bake_numbered_exercise/main.rb
```diff
@@ -3,18 +3,14 @@
...
+        V1.new.bake(exercise: exercise, number: number, suppress_solution_if: suppress_solution_if,
                     note_suppressed_solutions: note_suppressed_solutions)
       end
 
       def self.bake_solution_v1(exercise:, number:, divider: '. ')
         V1.new.bake_solution(exercise: exercise, number: number, divider: divider)
       end
-
-      def self.bake_even_solution_v1(exercise:, divider: '. ')
```
ajluk, line 14: I have also removed `bake_even_solution_v1` method, because it is not needed anymore after your changes :slightly_smiling_face: :

PR #244, opened 2021-05-26
extend chapter review container to accept other classes

kswanson33:


lib/kitchen/directions/chapter_review_container/v1.rb
```diff
@@ -4,10 +4,12 @@ module Kitchen::Directions::ChapterReviewContainer
...
     renderable
 
-    def bake(chapter:, metadata_source:)
+    def bake(chapter:, metadata_source:, klass: 'chapter-review')
       @metadata = metadata_source.children_to_keep.copy
+      @klass = klass
+      @title = klass == 'chapter-review' ? I18n.t(:eoc_chapter_review) : I18n.t(:"eoc.#{klass}")
       chapter.append(child: render(file: 'chapter_review.xhtml.erb'))
-      chapter.first('div.os-eoc.os-chapter-review-container')
+      chapter.element_children[-1]
```
brittany-johnson, line 12: Can you explain what this returns - I'm just curious

brittany-johnson (approved):

lib/kitchen/directions/chapter_review_container/v1.rb
kswanson33: This returns the container we just created. You may know this already, but `element_children` is a nokogiri function that returns all children that are elements (so, not text or newlines) as an array, & `[-1]` indexing returns the last item in an array. 

The previous implementation also tried to return the just-created container, but it did this by searching for the first instance matching the search query. This would stop working correctly if there were more than one of these containers on the chapter.

lib/kitchen/directions/chapter_review_container/v1.rb
```diff
@@ -4,10 +4,12 @@ module Kitchen::Directions::ChapterReviewContainer
   class V1
     renderable
 
-    def bake(chapter:, metadata_source:)
+    def bake(chapter:, metadata_source:, klass: 'chapter-review')
       @metadata = metadata_source.children_to_keep.copy
+      @klass = klass
+      @title = klass == 'chapter-review' ? I18n.t(:eoc_chapter_review) : I18n.t(:"eoc.#{klass}")
```
kswanson33, line 10: The ternary statement wouldn't be necessary if `Chapter Review` was stored under `eoc.chapter-review` & could be retrieved the same way as other titles, but I believe this would make this a breaking change. Thoughts?

KarinaMendez2: let's store this under eoc, so we can override the title from book specific translations

Issue #245, opened 2021-05-28
Link title for references from intorduction pages


ajluk:
In Sociology there are references present also in introduction pages: 
![image](https://user-images.githubusercontent.com/55203598/120009607-70233d00-bfdc-11eb-9470-0f1c5baaf1c7.png)
And currently we support only creating link title for references from non introduction pages: 
![image](https://user-images.githubusercontent.com/55203598/120010097-08b9bd00-bfdd-11eb-8007-18b69ef5a54f.png)
The diff between easybaked sociology 2e assembled file and kitchen baked in such case:
![image](https://user-images.githubusercontent.com/55203598/120010674-b62cd080-bfdd-11eb-95e7-0812b24ffc31.png)

Related main card: openstax/cnx-recipes#3504

ajluk:
The PR has been merged.

PR #246, opened 2021-05-28
create BakeChapterReferences v2

ajluk:
Issue: #245


LostMojo:
Did you want to merge this into main or bake-chapter-review-exercises-answer-key-strategy-sociology-recipe?

ajluk:
@LostMojo I was going to change the base branch when `bake-chapter-review-exercises-answer-key-strategy-sociology-recipe` will be merged into main.

lib/kitchen/directions/bake_chapter_references/v2.rb
```diff
@@ -0,0 +1,53 @@
...
+            <<~HTML
+              <a href="##{page.title.id}">
+                <h3 data-type="document-title" id="#{page.title.copied_id}">
+                  <span class="os-text" data-type="" itemprop="">#{page.title_text}</span>
+                </h3>
+              </a>
+            HTML
+          )
+        end
+      end
```
kswanson33, line 43: Instead of two loops, would it work to do something like

```ruby
chapter.pages.each do |page|
  # get the references and trash the h3, etc
  references.each do |reference|
    if page.is_introduction?
       # title a reference from an introduction page
    else
       # title a reference from a non introduction page
    end
  end
end
```

lib/kitchen/directions/bake_chapter_references/v2.rb
```diff
@@ -0,0 +1,53 @@
...
+  class V2
+    renderable
+
+    def bake(chapter:, metadata_source:, uuid_prefix: '.', klass: 'references')
+      @metadata = metadata_source.children_to_keep.copy
+      @klass = klass
+      @title = I18n.t(:references)
+      @uuid_prefix = uuid_prefix
+
+      chapter.references.search('h3').trash
```
kswanson33, line 14: is this line redundant with lines 19 & 31?

kswanson33 (changes requested):
If the only difference from v1 is baking references from introduction pages, I think this could be an argument to v1 instead of a v2. For example, `if page.is_introduction? && from_introduction`

lib/kitchen/directions/bake_chapter_references/v2.rb
ajluk: @kswanson33 As we discussed on slack the reason for creating two loops instead of one is that when the title is created for pages that are non introduction ones, the numbering of pages also include the intro ones, because they are all counted when they are in one loop:
image
image
So I'm leaving the two loops here because I couldn't found a solution how to fix this error occurring in one loop iteration.

lib/kitchen/directions/bake_chapter_references/v2.rb
ajluk: @kswanson33 the line was already there. Hmm maybe it is a leftover or we should keep it just for safety reasons? What do you think?

ajluk: @LostMojo through the lens I have found out that you added the line from above? Could you advise if it is still needed form some reason if you still remember off course :upside_down_face:

LostMojo: If I remember correctly, the one on line 13 was to catch h3's in references that didn't fall into a non-introduction page.

LostMojo (approved):
This should work with my recipe!

lib/kitchen/directions/bake_chapter_references/v1.rb
```diff
@@ -4,11 +4,12 @@ module Kitchen::Directions::BakeChapterReferences
   class V1
     renderable
 
-    def bake(chapter:, metadata_source:, uuid_prefix: '.', klass: 'references')
+    def bake(chapter:, metadata_source:, uuid_prefix: '.', klass: 'references', from_introduction: false)
       @metadata = metadata_source.children_to_keep.copy
       @klass = klass
       @title = I18n.t(:references)
       @uuid_prefix = uuid_prefix
+      @from_introduction = from_introduction
```
kswanson33, line 12: if this variable is not used in the template (and it doesn't look like it is), this line isn't necessary!

lib/kitchen/directions/bake_chapter_references/v1.rb
```diff
@@ -4,11 +4,12 @@ module Kitchen::Directions::BakeChapterReferences
...
 
       chapter.references.search('h3').trash
```
kswanson33, line 14: again regarding this line, along with line 20 & 34 ... i believe if you change this to `chapter.references.search('h3').each(&:trash)` it will trash all `h3` tags within a reference. if this is the behavior you're going for, i'd favor this approach over calling this line again on lines 20 & 34

kswanson33 (changes requested):
two very minor suggestions!

lib/kitchen/directions/bake_chapter_references/v2.rb
ajluk: Oh, Thank you MaryJo. :relaxed:

lib/kitchen/directions/bake_chapter_references/v1.rb
ajluk: Yes, this gives the expected result. Thank you Kendra! :slightly_smiling_face:

lib/kitchen/directions/bake_chapter_references/v2.rb
ajluk: @LostMojo can I have just one more question regarding references in your book because as I understood from your answer from above is that you needed to remove the h3 titles from references from intro pages but you were not creating a k\link title for them, is that the case?

kswanson33 (approved):

lib/kitchen/directions/bake_chapter_references/v2.rb
LostMojo: They weren't in the introduction pages, they were composite pages (that don't get picked up as real pages, sometimes). I don't... think I was creating a k/link title for them, though :\

lib/kitchen/directions/bake_chapter_references/v1.rb
```diff
@@ -4,28 +4,51 @@ module Kitchen::Directions::BakeChapterReferences
   class V1
     renderable
 
-    def bake(chapter:, metadata_source:, uuid_prefix: '.', klass: 'references')
+    def bake(
+      chapter:,
+      metadata_source:,
+      uuid_prefix: '.',
+      klass: 'references',
+      from_introduction: false
```
jpslav, line 12: Assuming that we do always want to bake references in introductions if they are there (discussion[https://openstax.slack.com/archives/C022NETP39N/p1623768035286400]), let's remove this flag and then change the block of code starting on line 21 (from `chapter.non_introduction_pages...`) to

```ruby
chapter.pages.each do |page|
  references = page.references        
  next if references.none?

  if page.is_introduction?
    title =
      <<~HTML
        <a href="##{page.title.id}">
          <h3 data-type="document-title" id="#{page.title.copied_id}">
            <span class="os-text" data-type="" itemprop="">#{page.title_text}</span>
          </h3>
        </a>
      HTML
  else
    references.search('h3').trash        
    title = Kitchen::Directions::EocSectionTitleLinkSnippet.v1(page: page)        
  end

  references.each do |reference|          
    reference.prepend(child: title)        
  end
end
```

Then we can remove the block starting on line 32.

lib/kitchen/directions/bake_chapter_references/v1.rb
```diff
@@ -4,28 +4,51 @@ module Kitchen::Directions::BakeChapterReferences
...
+    )
       @metadata = metadata_source.children_to_keep.copy
       @klass = klass
       @title = I18n.t(:references)
       @uuid_prefix = uuid_prefix
 
-      chapter.references.search('h3').trash
+      chapter.references.search('h3').each(&:trash)
```
jpslav, line 19: Did this not work before?  The `trash` method on the enumerator returned by `search` should do exactly this. https://github.com/openstax/kitchen/blob/main/lib/kitchen/element_enumerator_base.rb#L371-L373

lib/kitchen/directions/bake_chapter_references/v1.rb
```diff
@@ -4,28 +4,51 @@ module Kitchen::Directions::BakeChapterReferences
...
-        references.search('h3').trash
         title = Kitchen::Directions::EocSectionTitleLinkSnippet.v1(page: page)
 
         references.each do |reference|
           reference.prepend(child: title)
         end
       end
 
-      @content = chapter.pages.references.cut.paste
+      if from_introduction == true
```
jpslav, line 32: per above, this block of code is no longer required.  But for the future, this line can instead simply be

```ruby
if from_introduction
```

you don't have to compare it to `true` because `from_introduction` is already a boolean.

jpslav (changes requested):

lib/kitchen/directions/bake_chapter_references/v1.rb
jpslav:
```ruby
def bake
  ...
  bake_page_references(page: chapter.introduction_page)

    bake_page_references(page: page)
  end
  ...
end


def bake_page_references(page:)
  return if page.nil?

  references = page.references        
  return if references.none?

  if page.is_introduction?
    title =
      <<~HTML
        <a href="##{page.title.id}">
          <h3 data-type="document-title" id="#{page.title.copied_id}">
            <span class="os-text" data-type="" itemprop="">#{page.title_text}</span>
          </h3>
        </a>
      HTML
  else
    references.search('h3').trash        
    title = Kitchen::Directions::EocSectionTitleLinkSnippet.v1(page: page)        
  end

  references.each do |reference|          
    reference.prepend(child: title)        
  end
end
```

jpslav (approved):

PR #249, opened 2021-06-01
Rm table summary attr in numbered tables

brittany-johnson:
[Conversation and decision here ](https://openstax.slack.com/archives/C022NETP39N/p1621960840206100)
Currently, unnumbered tables did not have the summary attribute but numbered tables should also not have table attribute anymore

### DIFF
--- a/lib/kitchen/directions/bake_numbered_table/v1.rb
+++ b/lib/kitchen/directions/bake_numbered_table/v1.rb
@@ -26,7 +26,6 @@ module Kitchen::Directions::BakeNumberedTable
       caption_title = ''
 
       if (title = table.first("span[data-type='title']")&.cut)
-        # new_summary += title.text
         caption_title = <<~HTML
           \n<span class="os-title" data-type="title">#{title.children}</span>
         HTML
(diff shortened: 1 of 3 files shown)


lib/kitchen/directions/bake_numbered_table/v1.rb
```diff
@@ -22,21 +22,17 @@ def bake(table:, number:, always_caption: false)
       table.parent.add_class('os-column-header-container') if table.column_header?
 
       # TODO: extra spaces added here to match legacy implementation, but probably not meaningful?
-      new_summary = "#{table_label} "
       new_caption = ''
       caption_title = ''
 
       if (title = table.first("span[data-type='title']")&.cut)
-        new_summary += title.text
+        # new_summary += title.text
```
kswanson33, line 29: don't forget to take this out!

LostMojo, line 29: if it's gone-- delete the comment?

kswanson33 (approved):
Looks good, there's one stray commented line but otherwise this seems solid

LostMojo (changes requested):
One small change, otherwise, looks good!

lib/kitchen/directions/bake_numbered_table/v1.rb
LostMojo: whoops, ditto

LostMojo (approved):

lib/kitchen/directions/bake_numbered_table/v1.rb
brittany-johnson: OOPS 
thank you

brittany-johnson: thank you!

PR #252, opened 2021-06-03
num tables and preface title update

brittany-johnson:
Remove summary attribute from numbered tables, add option to pass title element to `BakePreface`


lib/kitchen/directions/bake_numbered_table/v1.rb
```diff
@@ -4,6 +4,7 @@ module Kitchen::Directions::BakeNumberedTable
   class V1
 
     def bake(table:, number:, always_caption: false)
+      table.remove_attribute('summary')
```
brittany-johnson, line 7: [in this PR](https://github.com/openstax/kitchen/pull/249) I thought I had removed the summary attr BUT it was still there but with different data

lib/kitchen/directions/bake_preface/main.rb
```diff
@@ -3,8 +3,8 @@
 module Kitchen
   module Directions
     module BakePreface
-      def self.v1(book:)
-        V1.new.bake(book: book)
+      def self.v1(book:, title_element:)
```
brittany-johnson, line 6: add an option because am gov is a div instead of h1

LostMojo: is there a reason to define them seperately rather than giving the second argument a default?

LostMojo (commented):
Loos good! I only have one question about defining two v1 methods-- I've been doing it by giving the second argument a default value, but is there a benefit to defining it twice like this?

LostMojo (approved):

PR #253, opened 2021-06-03
Making a default strategy for eob exercises

KarinaMendez2:


kswanson33 (approved):

lib/kitchen/directions/move_solutions_to_answer_key/v1.rb
```diff
@@ -11,6 +11,8 @@ def bake(chapter:, metadata_source:, strategy:, append_to:)
           Strategies::UPhysics
         when :american_government
           Strategies::AmericanGovernment
```
kswanson33, line 13: this may be why codecov is complaining ... does this strategy exist anymore? if not, this case is not needed

kswanson33 (changes requested):
oh wait i just saw that codecov is complaining nvm

brittany-johnson (approved):

ajluk (approved):

kswanson33 (approved):

PR #256, opened 2021-06-07
stop removing the first strong tag from notes

kswanson33:


KarinaMendez2 (approved):

brittany-johnson (approved):

lib/kitchen/directions/bake_notes/bake_numbered_notes/main.rb
```diff
@@ -16,7 +16,6 @@ def self.bake_note_exercise(note:, exercise:, divider: ' ')
         exercise.add_class('unnumbered')
         # bake problem
         exercise.problem.wrap_children('div', class: 'os-problem-container')
-        exercise.problem.first('strong')&.trash
```
brittany-johnson, line 19: I approved but realized after I have a question lol 

Is it that all numbered notes should keep the strong tag - so this doesn't need to be optional?

kswanson33: `BakeNumberedNotes` is supposed to automatically generate the note title, but some books also have the title added manually in the content, so there ends up being a duplicate that needs to be removed. however, some other books don't have this manual title. since there's no special selector to differentiate the title from other `<strong>` content that we do want to keep, this has ended up deleting necessary content. so the new plan is to get the CMs to remove the duplicate titles from the content, and therefore removing the need for this line.

kswanson33: does that ... make sense

brittany-johnson: Ahhh! Yes I see, thanks for the explanation!

PR #257, opened 2021-06-07
delete abstract and description from preface

kswanson33:


lib/kitchen/directions/bake_preface/v1.rb
```diff
@@ -4,6 +4,7 @@ module Kitchen::Directions::BakePreface
   class V1
     def bake(book:, title_element:)
       book.pages('$.preface').each do |page|
+        page.search('div[data-type="description"], div[data-type="abstract"]').each(&:trash)
```
brittany-johnson, line 7: Is this going to negatively affect other books currently using v1?

kswanson33: good question... idk lol 😬

kswanson33: so I double checked chem and uphysics, and neither of them will be impacted. i also created another test case/extended the case to verify that this doesn't make `BakePreface` blow up without these elements. but i don't know about other books.

brittany-johnson: I just check Am Gov and doesn't look like it'll be affected either. 

But this does give me the idea that maybe we should be running recipe spec as part of github checks in kitchen repo?

brittany-johnson (approved):

Issue #259, opened 2021-06-08
Use twitter sorting in Key Terms


wilkus27:
Terms in Key Terms should be sorted with rules from language of the book.
We should use for that `twitter_cldr` library. Similar work was done for indexes: https://github.com/openstax/kitchen/pull/240

wilkus27:
Terms with polish diacritics are sorted properly:

_Key Terms, Chapter 2_

![image](https://user-images.githubusercontent.com/40228854/123103079-f5f8a380-d435-11eb-8f94-0db2fefda364.png)

PR #260, opened 2021-06-08
Sort key terms using twitter

wilkus27:
Issue: https://github.com/openstax/kitchen/issues/259


lib/kitchen/directions/bake_chapter_glossary/v1.rb
```diff
@@ -12,8 +12,13 @@ def bake(chapter:, metadata_source:, append_to: nil, uuid_prefix: '')
...
       return if definitions.none?
-      definitions.sort_by! do |definition|
-        [definition.first('dt').text.downcase, definition.first('dd').text.downcase]
+
+      definitions.sort! do |def1, def2|
+        if def1.first('dt').text.downcase == def2.first('dt').text.downcase
+          I18n.sort_strings(def1.first('dd').text.downcase, def2.first('dd').text.downcase)
+        else
+          I18n.sort_strings(def1.first('dt').text.downcase, def2.first('dt').text.downcase)
+        end
```
jpslav, line 21: I see how you're working to make sorting use `I18n.sort_strings`.  When I look at this, I see a lot of repetition.  For example, 6 calls to `.text.downcase`, 6 calls to `.first(blah)`.  As you and the other developers get more senior, one goal is to recognize repetition like this and ask yourself (or your teammates) if there's a way to "DRY it up".  DRY = Don't Repeat Yourself.  

A maybe-too-aggressive way to DRY this up would be to override all String sorting by monkey patching the `String` class:

```ruby
class String
  def <=>(other)
    I18n.sort_strings(self, other)
  end
end
```

This may lead to unintended consequences elsewhere and will also be less efficient than the normal [C implementation of `<=>`](https://ruby-doc.org/core-2.6/String.html#method-i-3C-3D-3E).

Another more limited way would be to provide a helper method on `ElementBase`:

```ruby
class ElementBase
  def i18n_sortable_text
    I18nSortableText.new(text)
  end
```

Where you've added this new class to `kitchen/lib`:

```ruby
class I18nSortableText
  attr_reader :text

  def initialize(text)
    @text = text
  end

  def <=>(other)
    I18n.sort_strings(text.downcase, other.text.downcase)
  end
end
```

and then your `BakeChapterGlossary::V1` code becomes

```ruby
definitions.sort_by! do |definition|        
  [definition.first('dt').i18n_sortable_text, definition.first('dd').i18n_sortable_text]
end
```

A third way is to do what we did in `BakeIndex`, that is make a `Definition` class here that is given an element and can implement its own sorting method.

CHANGELOG.md
```diff
@@ -6,6 +6,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+* Sort terms in `BakeChapterGlossary` in language specific way (minor)
```
jpslav, line 9: If this PR changes the sort order for others, the change should be major.

jpslav (changes requested):

wilkus27:
@jpslav I decided to try to apply third way that you suggested. PR is ready for rereview.

lib/kitchen/directions/bake_chapter_glossary/v1.rb
```diff
@@ -4,21 +4,85 @@ module Kitchen::Directions::BakeChapterGlossary
   class V1
     renderable
 
+    class SortableDefinition
```
jpslav, line 7: There's nothing about this class that is specific to definitions.  So I'd propose making the name of it be more generic.  What about moving this class to `lib/kitchen`:

```ruby
class I18nString < String
  def <=>(other)
    I18n.sort_strings(self, other)
  end
end
```

lib/kitchen/directions/bake_chapter_glossary/v1.rb
```diff
@@ -4,21 +4,85 @@ module Kitchen::Directions::BakeChapterGlossary
...
+      # Text from tag dt
+      attr_reader :term_text
+
+      # Text from tag dd
+      attr_reader :description_text
+
+      # Clipboard with content of dl tag
+      attr_reader :def_content
+
+      def initialize(term_text:, description_text:, def_content:)
```
jpslav, line 30: since `def_content` is an element, let's call it `element` instead.  And actually, since the other two arguments are included in `def_content`, let's pull them out inside:

```ruby
def initialize(element)
  @term_text = I18nString.new(element.first('dt').text.downcase)
  @description_text = I18nString.new(element.first('dd').text.downcase)
  @element = element
end
```

Furthermore, `term_text` and `description_text` aren't really used except for sorting.  So I'd recommend the following:

```ruby
class Definition
  attr_reader :element

  def initialize(element)
    term = I18nString.new(element.first('dt').text.downcase)
    description = I18nString.new(element.first('dd').text.downcase)
    @sortable = [term, definition]
    @element = element
  end

  def <=>(other)
    sortable <=> other.sortable
  end
 
  protected

  attr_reader :sortable
end
```

Protected attributes are readable by other instances of the same class.  No other class needs this information.

lib/kitchen/directions/bake_chapter_glossary/v1.rb
```diff
@@ -4,21 +4,85 @@ module Kitchen::Directions::BakeChapterGlossary
...
+        self
+      end
+    end
+
     def bake(chapter:, metadata_source:, append_to: nil, uuid_prefix: '')
       @metadata = metadata_source.children_to_keep.copy
       @klass = 'glossary'
       @title = I18n.t(:eoc_key_terms_title)
       @uuid_prefix = uuid_prefix
+      @glossary = Glossary.new
```
jpslav, line 70: Because we added the spaceship operator to the `Definition` class, I think you can get rid of the `Glossary` class and simply make it an array:

```ruby
@glossary = []
...
@glossary.push(Definition.new(definition_element.cut))
```

lib/kitchen/directions/bake_chapter_glossary/v1.rb
```diff
@@ -4,21 +4,85 @@ module Kitchen::Directions::BakeChapterGlossary
...
+            description_text: definition_element.first('dd').text.downcase,
+            def_content: definition_element.cut
+          )
+        )
       end
 
       chapter.glossaries.trash
 
-      @content = definitions.paste
+      @content = render(file: 'v1.xhtml.erb')
```
jpslav, line 85: You can leave this as is, but there's not a lot in this `.xhtml.erb` file.  I think you can delete the template file and replace it with this one line of code.

```ruby
@content = @glossary.sort.map { |definition| definition.element.paste }
```

jpslav (changes requested):

lib/kitchen/directions/bake_chapter_glossary/v1.rb
wilkus27: I tried to apply this but it returns an array with `dl` elements:

image

jpslav: Sorry, add `.join` at the end

```ruby
@content = @glossary.sort.map { |definition| definition.element.paste }.join
```

which concatenates all items in the array into one string

wilkus27: Thank you! It's working working now as intended :slightly_smiling_face:

jpslav (approved):

PR #262, opened 2021-06-09
Solution vs solutions

kswanson33:


kswanson33:
switching from v1/v2 approach to passing in `solutions_plural: true/false` method arg

lib/kitchen/directions/move_solutions_to_answer_key/v1.rb
```diff
@@ -19,7 +19,8 @@ def bake(chapter:, metadata_source:, strategy:, append_to:, strategy_options: {}
 
       append_to.append(child:
         <<~HTML
-          <div class="os-eob os-solutions-container" data-type="composite-page" data-uuid-key=".solutions#{chapter.count_in(:book)}">
+          <div class="os-eob os-solution#{'s' if solutions_plural}-container" data-type="composite-page" \
+          data-uuid-key=".solution#{'s' if solutions_plural}#{chapter.count_in(:book)}">
```
ajluk, line 23: @kswanson33 should `@solutions_or_solution` replace the condition here as well?

ajluk (approved):

PR #265, opened 2021-06-09
Bake Numbered Notes V3

LostMojo:
A V3 of bake numbered notes that grab their number from the exercise before them. Also adds a `#previous` method to note_element to more easily access the immediately previous element


lib/kitchen/directions/bake_notes/bake_numbered_notes/main.rb
```diff
@@ -11,8 +11,12 @@ def self.v2(book:, classes:)
         V2.new.bake(book: book, classes: classes)
       end
 
-      # Used by V1, V2
-      def self.bake_note_exercise(note:, exercise:, divider: ' ')
+      def self.v3(book:, classes:, suppress_solution: true)
```
jpslav, line 14: Like I think @KarinaMendez2 or someone said the other day, can you write docs inline here that say what is different about v3?

lib/kitchen/directions/bake_notes/bake_numbered_notes/main.rb
```diff
@@ -21,8 +25,14 @@ def self.bake_note_exercise(note:, exercise:, divider: ' ')
         return unless exercise.solution
 
         # bake solution in place
-        BakeNumberedExercise.bake_solution_v1(
-          exercise: exercise, number: note.first('.os-number').text.gsub(/#/, ''), divider: divider)
+        if suppress_solution
+          exercise.add_class('os-hasSolution')
```
jpslav, line 29: do we not add this class if there is a solution but `suppress_solution` is false?

lib/kitchen/element_base.rb
```diff
@@ -508,6 +508,22 @@ def parent
...
+    def previous
+      node = raw.previous
+      return if node.nil?
+
+      sibling = Element.new(
+        node: node,
+        document: document,
+        short_type: "previous(#{short_type})"
+      )
+      sibling.to_s =~ /\S/ ? sibling : sibling.previous
```
jpslav, line 524: What is this line doing?  Looking backwards until it finds a non whitespace element sibling?

If so, probably more efficient to do that at the start of the method so we don't make a bunch of unused `Element` objects.  Actually can probably add to `lib/kitchen/patches/nokogiri.rb`:

```ruby
module Nokogiri
  class Node
    def previous_element
      prev = previous
      return nil if prev.nil?

      prev.text? ? prev.previous_element : prev
    end
  end
end
```

then change your first line above to 

```ruby
node = raw.previous_element
```

and ditch the last line

jpslav (changes requested):
Didn't look too much at the guts of the baking, more the previous stuff

lib/kitchen/element_base.rb
jpslav: Note this is also more efficient b/c it isn't serializing the element to string a lot

lib/kitchen/directions/bake_notes/bake_numbered_notes/main.rb
LostMojo: if there is a solution, `BakeNumberedExercise.bake_solution_v1` adds that class to the exercise. If the solution is suppressed, it apparently still needs the `os-hasSolution` class

ajluk: @jpslav @LostMojo I have made some changes (together with you suggestions JP) in this PR: https://github.com/openstax/kitchen/pull/181. Please look into it if you would like to.

jpslav (approved):

Issue #266, opened 2021-06-10
Unnumbered EOB References


wilkus27:
Move sections with class `references` to EOB. They should be prepend with chapter title:

![image](https://user-images.githubusercontent.com/40228854/121498731-0001b780-c9dd-11eb-917f-750fb9fd4f4b.png)

PR #267, opened 2021-06-10
Unnumbered eob references

wilkus27:
Issue: https://github.com/openstax/kitchen/issues/266


lib/kitchen/composite_page_element.rb
```diff
@@ -43,13 +43,21 @@ def is_index?
...
     end
 
+    # Returns true if this page is a book section reference
+    #
+    # @return [Boolean]
+    #
+    def is_section_reference?
+      has_class?('os-references-container')
+    end
+
```
wilkus27, line 61: I checked and class `reference` is used in books that contain references in form of citation, similar to footnotes:

image

While class `references` is used when book contain a whole section in Introduction page:

image

brittany-johnson: The one letter difference is so unfortunate 😂

brittany-johnson: Maybe it would be good to add the difference that you explained in this comment to the file

brittany-johnson (approved):
Looks good!

kswanson33 (approved):

lib/kitchen/composite_page_element.rb
wilkus27: Yes, sure :slightly_smiling_face: It will be less confusing in the future. I added needed comment.

PR #268, opened 2021-06-10
v2 for numbered table captions

kswanson33:


kswanson33:
I chose to use a V2 instead of adding a method argument option. Shared code is still in V1. Thoughts on if shared code should be moved to main? The advantage of keeping it in V1 is that it keeps main from getting cluttered, especially if we need to make another version of the shared code at some point.

lib/kitchen/directions/bake_numbered_table/v2.rb
```diff
@@ -0,0 +1,30 @@
+# frozen_string_literal: true
+
+module Kitchen::Directions::BakeNumberedTable
+  # Difference from v1: only in the caption
```
ajluk, line 4: I really like that you have added the comment about the versions differences. It makes the review and use of the version later in the recipes much easier :heart: Maybe adding what is the difference in caption exactly could be useful (that a caption has a title already which needs to be putted in `span.os-caption` if I'm right here :slightly_smiling_face: )?

spec/directions/bake_example_spec.rb
```diff
@@ -99,6 +101,58 @@
...
+            <div class="body">
+              <h4 data-type="title">example title becomes h4</h4>
+              <p>content</p>
+              #{table}
+            </div>
+          </div>
+        HTML
+      )
+    end
+  end
```
ajluk, line 154: I'm not quite understand how this part works is it checking if the table is placed exactly in the example? :sweat_smile:

ajluk (commented):
@kswanson33 I have only left some questions/comments, but everything looks good here. Regarding your question from above if the method `bake_table` should be moved to `main` file I'm not sure either. As you have written above there are some advantages and disadvantages  of this step, but the part of code you have putted in this method will be used again I suppose by different versions if there will be more of them, so maybe we could move such methods to another file and use it in the versions where we want to, but it's just a thought.

lib/kitchen/directions/bake_numbered_table/v2.rb
kswanson33: makes sense! I wasn't sure how to explain the difference verbally, but I'll give it a try

spec/directions/bake_example_spec.rb
kswanson33: there's a line in `BakeExample` that converts a `data-type=["title"]` span to an h4. But this shouldn't impact other elements contained within the example. It was impacting my tables, so I changed the line & added this spec to make sure 😄

lib/kitchen/directions/bake_numbered_table/v2.rb
ajluk: Maybe something similar to what I have written above f.ex  Difference from v1: only in the caption with title which needs to put in span.os-caption`. It's not pretty, but does the job I suppose.

lib/kitchen/directions/bake_example.rb
```diff
@@ -21,7 +21,10 @@ def self.v1(example:, number:, title_tag:)
                .store("#{I18n.t(:example_label)} #{number}", label: example.id)
 
         example.titles.each do |title|
-          next if title.parent.has_class?('os-caption-container')
+          if title.parent.has_class?('os-caption-container') || \
```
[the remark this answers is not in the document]
ajluk: Oh, that's an interesting hack!

ajluk (approved):

PR #269, opened 2021-06-10
Allow BakeFootnotes to number footnotes using roman numerals

kswanson33:


lib/kitchen/directions/bake_footnotes/main.rb
```diff
@@ -6,6 +6,41 @@ module BakeFootnotes
       def self.v1(book:)
```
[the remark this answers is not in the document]
kswanson33: There's an additional difference in how footnotes are numbered: instead of 

```ruby
book.chapters.each do |chapter|
  bake_footnotes_within(chapter)
end
```

as in v1, v2 uses

```ruby
bake_footnotes_within(book.chapters)
```

For this difference, I could pass in a variable like `within: :book` or `within: :chapter` and then do something like

```ruby
case within
when :book
  bake_footnotes_within(book.chapters)
when :chapter
  # iteration to bake by chapter
end
```

It was at this point that I started to feel like the divergence could be sufficient for a v2. But v1 is obviously still possible.

Regarding the roman numeral argument, only the top-level page footnotes are numbered with roman numerals, while the footnotes within chapters are numbered normally. Would `number_format_pages` work for the variable name instead, leaving it open to separately specify the chapter numbering with `number_format_chapters` later?

lib/kitchen/directions/bake_footnotes/v1.rb
```diff
@@ -10,32 +10,12 @@ def bake(book:)
       book.body.element_children.only(Kitchen::PageElement,
                                       Kitchen::CompositePageElement,
                                       Kitchen::CompositeChapterElement).each do |page|
-        bake_footnotes_within(page)
+        Kitchen::Directions::BakeFootnotes.bake_footnotes_within(page)
```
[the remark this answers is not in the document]
kswanson33: i double checked, and it does seem to be necessary. Using just `bake_footnotes_within` makes this look for a function within `V2` of that name, while using `BakeFootnotes.bake_footnotes_within` makes it look for `...::V2::BakeFootnotes.etc`. Same for `Directions::BakeFootnotes.etc`

lib/kitchen/directions/bake_footnotes/main.rb
```diff
@@ -6,6 +6,31 @@ module BakeFootnotes
...
         V1.new.bake(book: book)
       end
+
+      def self.v2(book:)
+        V2.new.bake(book: book)
+      end
+
+      def self.bake_footnotes_within(container, number_format: :arabic)
```
jpslav, line 14: Since we're stuck writing `Kitchen::Directions::BakeFootnotes::this_method_name` can we take `bake_footnotes` out of the name?  Too repetitive to say `Kitchen::Directions::BakeFootnotes::bake_footnotes...`

Maybe `def self.within(container, number_format: :arabic)` ?

lib/kitchen/directions/bake_footnotes/v1.rb
```diff
@@ -10,32 +10,12 @@ def bake(book:)
       book.body.element_children.only(Kitchen::PageElement,
                                       Kitchen::CompositePageElement,
                                       Kitchen::CompositeChapterElement).each do |page|
```
jpslav, line 12: just checking that composite chapters are meant to be included? since the variable name is `page`

lib/kitchen/directions/bake_footnotes/v2.rb
```diff
@@ -0,0 +1,18 @@
...
+
+module Kitchen::Directions::BakeFootnotes
+  # Difference from v1:
+  # 1. top-level pages are numbered with roman numerals
+  # 2. footnote counts don't reset with each new chapter
+  class V2
+    def bake(book:)
+      book.body.element_children.only(Kitchen::PageElement,
+                                      Kitchen::CompositePageElement,
+                                      Kitchen::CompositeChapterElement).each do |page|
```
jpslav, line 11: You may not need `Kitchen::` on these three lines, ditto V1.  They aren't needed when used in BakeToc this way.

lib/kitchen/patches/integer.rb
```diff
@@ -0,0 +1,22 @@
+# frozen_string_literal: true
+
+# Monkey patches for +Integer+
+#
+class Integer
+  ROMAN_NUMERALS = %w[0 i ii iii iv v vi vii viii ix x].freeze
```
jpslav, line 6: How likely will we go beyond 10?

lib/kitchen/patches/integer.rb
```diff
@@ -0,0 +1,22 @@
...
+  #
+  def to_format(format)
+    case format
+    when :arabic
+      to_s
+    when :roman
+      raise 'Unknown conversion to Roman numerals' if self >= ROMAN_NUMERALS.size
+
+      ROMAN_NUMERALS[self]
+    end
```
jpslav, line 20: can you add an `else` to the `case` to `raise` when `format` is unknown?

jpslav (changes requested):

lib/kitchen/patches/integer.rb
kswanson33: extended to 20 just in case, but I think this is unlikely

jpslav (approved):

PR #272, opened 2021-06-14
Fix for index when words start with number

KarinaMendez2:
From easy-baked, the behavior for words like `Ötzi the Iceman` are expected to be grouped as symbols, but the currently expected behavior is for those to be grouped by letter, this PR changes fixes that issue.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,6 +31,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 and from v2 the lack of additional "os-section-area" and os-#{@klass} wrapper (minor)
 * Add a condition in BakeNumberedExercise to make it possible to suppress even solutions in the Answer Key (minor)
 * Fix BakeFurtherResearch baking with main bake script error by breaking the loop if further research sections are not present (minor)
+* Fix for `BakeIndex` for words that start with a number (minor)
 
 ## [5.0.0] - 2021-06-02
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,7 +31,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 and from v2 the lack of additional "os-section-area" and os-#{@klass} wrapper (minor)
 * Add a condition in BakeNumberedExercise to make it possible to suppress even solutions in the Answer Key (minor)
 * Fix BakeFurtherResearch baking with main bake script error by breaking the loop if further research sections are not present (minor)
-* Fix for `BakeIndex` for words that start with a number (minor)
+* Fix for `BakeIndex` for words that start with a number (major)
 
 ## [5.0.0] - 2021-06-02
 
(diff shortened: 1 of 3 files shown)


kswanson33 (approved):

lib/kitchen/directions/bake_index/v1.rb
```diff
@@ -137,7 +137,7 @@ def bake(book:)
 
     def add_term_to_index(term_element, page_title)
       group_by = term_element.text.strip[0]
-      group_by = I18n.t(:eob_index_symbols_group) unless group_by.match?(/\w/)
+      group_by = I18n.t(:eob_index_symbols_group) unless group_by.match?(/[a-zA-Z]/)
```
jpslav, line 140: I think this will need to handle words that start with accent marks too?  

Can you change line 139 to be:

```ruby
group_by = I18n.transliterate(term_element.text.strip[0])
```

to remove accent marks on the group by? (https://stackoverflow.com/a/15696883) And also add a spec that a word starting with an accent mark gets grouped appropriately? (e.g. "éblahblah")

jpslav (changes requested):

spec/directions/bake_index/v1_spec.rb
```diff
@@ -174,6 +174,24 @@
     expect(a_section.items.first.term_text).to eq 'temperature'
   end
 
+  it 'groups by number when word starts with a number' do
+    a_section.add_term(text_only_term('70S ribosome'))
+    a_section.add_term(text_only_term('3-PGA'))
+    a_section.add_term(text_only_term('5’ cap'))
+    a_section.add_term(text_only_term('3-phosphoglycerate'))
+
+    expect(a_section.items.map(&:term_text)).to eq ['3-PGA', '3-phosphoglycerate', '5’ cap', '70S ribosome']
```
jpslav, line 183: This is testing the sort, not the grouping.  Ditto below.  You might need to add number and accent mark terms into the larger bake spec to see that they go where they should

lib/kitchen/directions/bake_index/v1.rb
```diff
@@ -136,8 +136,9 @@ def bake(book:)
     end
 
     def add_term_to_index(term_element, page_title)
-      group_by = term_element.text.strip[0]
-      group_by = I18n.t(:eob_index_symbols_group) unless group_by.match?(/\w/)
+      # group_by = term_element.text.strip[0]
```
jpslav, line 139: Remove commented code

CHANGELOG.md
```diff
@@ -31,6 +31,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 and from v2 the lack of additional "os-section-area" and os-#{@klass} wrapper (minor)
 * Add a condition in BakeNumberedExercise to make it possible to suppress even solutions in the Answer Key (minor)
 * Fix BakeFurtherResearch baking with main bake script error by breaking the loop if further research sections are not present (minor)
+* Fix for `BakeIndex` for words that start with a number (minor)
```
jpslav, line 34: This is "major" as it changes the output

Remember that a one-character change could be "major".

jpslav (changes requested):

jpslav (approved):

PR #273, opened 2021-06-14
Update Bake Summary to skip pages without a summary

LostMojo:
catches a 'no summary on page' error and moves to the next page. (had a chat with Larissa, said yes)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 * Allow `BakeChapterSummary` to skip pages where there is no summary (minor)
+* Change `PageElement#summary` to return nil instead of raise an error if no matches (major?)
+* Add `PageElement#summary!` method that raises an error if no matches (minor)
 
 * Add Rubocop and a working CHANGELOG check to GitHub actions (patch)
 * Allow `BakeFootnotes` to number footnotes with Roman numerals (minor)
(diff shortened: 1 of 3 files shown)


lib/kitchen/directions/bake_chapter_summary.rb
```diff
@@ -26,7 +26,11 @@ def bake(chapter:, metadata_source:, uuid_prefix: '.', klass: 'summary')
 
           # TODO: include specific page types somehow without writing it out
           chapter.non_introduction_pages.each do |page|
-            summary = page.summary
+            begin
+              summary = page.summary
+            rescue
+              next
+            end
```
LostMojo, line 33: Is this the best way to do this? #summary raises a Recipe error when it can't find anything, but the other search options seemed like they'd eat up a lot of time and might not use the right selectors.

kswanson33: `next unless page&.summary` would also accomplish this i believe

LostMojo: That only catches it if page is nil-- #summary raises it's own error by using #first!

CHANGELOG.md
```diff
@@ -6,6 +6,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+* Allow `BakeChapterSummary` to skip pages where there is no summary (minor)
+* Change `PageElement#summary` to return nil instead of raise an error if no matches (major?)
```
kswanson33, line 10: I believe this is still minor ... correct me if I'm wrong tho

lib/kitchen/directions/bake_chapter_summary.rb
```diff
@@ -27,6 +27,9 @@ def bake(chapter:, metadata_source:, uuid_prefix: '.', klass: 'summary')
           # TODO: include specific page types somehow without writing it out
           chapter.non_introduction_pages.each do |page|
             summary = page.summary
+
+            next if summary.nil?
```
kswanson33, line 31: this works, but you could also use `next unless summary` to check for truthiness

kswanson33 (approved):
minor changes suggested, but not essential!

PR #277, opened 2021-06-15
Bake Inline Lists

LostMojo:
inserts an inline-list-separator between elements in an inline list

### DIFF
--- a/lib/kitchen/directions/bake_inline_lists.rb
+++ b/lib/kitchen/directions/bake_inline_lists.rb
@@ -1,0 +1,22 @@
+# frozen_string_literal: true
+
+module Kitchen
+  module Directions
+    # Bakes inline lists with the desired list separator
+    # Does not separate the last list item
+    #
+    module BakeInlineLists
+      def self.v1(book:)
+        list_separator = '; '
+        separator_class = '-os-inline-list-separator'
+
+        inline_lists = book.search('span[data-display="inline"][data-type="list"]')
+        inline_lists.each do |list|
+          list.search('span[data-type="item"]')[0..-2].each do |item|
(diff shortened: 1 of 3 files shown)


lib/kitchen/directions/bake_inline_lists.rb
```diff
@@ -0,0 +1,22 @@
...
+  module Directions
+    # Bakes inline lists with the desired list separator
+    # Does not separate the last list item
+    #
+    module BakeInlineLists
+      def self.v1(book:)
+        list_separator = '; '
+        separator_class = '-os-inline-list-separator'
+
+        inline_lists = book.search('span[data-display="inline"][data-type="list"]')
```
LostMojo, line 13: Is doing the entire book at once okay? It doesn't need any chapter data, and doesn't rely on/break any order of operations of the rest of the recipe.

LostMojo: This might've just slowed down the recipe by a lot. New plan

lib/kitchen/directions/bake_inline_lists.rb
```diff
@@ -0,0 +1,22 @@
...
+
+module Kitchen
+  module Directions
+    # Bakes inline lists with the desired list separator
+    # Does not separate the last list item
+    #
+    module BakeInlineLists
+      def self.v1(book:)
+        list_separator = '; '
+        separator_class = '-os-inline-list-separator'
```
jpslav, line 11: since these strings don't change, make them constants...

```ruby
module BakeInlineLists
  LIST_SEPARATOR = '; '
  SEPARATOR_CLASS = '-os-inline-list-separator'

  def self.v1(book:)
```

jpslav (changes requested):

PR #280, opened 2021-06-17
Direction for a generic eoc section

kswanson33:


lib/kitchen/directions/bake_generic_eoc_section/v1.rb
```diff
@@ -0,0 +1,37 @@
...
+      @uuid_prefix = uuid_prefix
+
+      section_clipboard = Kitchen::Clipboard.new
+
+      chapter.non_introduction_pages.each do |page|
+        sections = page.search("section.#{klass}")
+        # Transforms the content of the section according to the given strategies, if given
+        content_transform.each do |strat|
+          sections.each do |section|
+            Kitchen::Directions::EocContentTransform.v1(section: section, strategy: strat)
```
kswanson33, line 20: What do y'all think of this way of addressing content changes to EoC sections? There are a couple of different ways of invoking methods in ruby, and I chose to create a mini-direction, `EocContentTransform`, whose methods perform small modifications to `<section>` elements. But I think this could be done in other ways too. Thoughts?

kswanson33 (commented):
An example of how this would be called in recipes, using Key Terms (for no particular reason):

```ruby
BakeGenericEocSection.v1(chapter: chapter, metadata_source: metadata, klass: 'key-terms', 
                   content_transform: [:remove_title, :key_terms])
```

The above would both remove a spare title from the section and bake any key terms in that section (assuming a `key_terms` method was defined and specced in EocContentTransform). That way, directions can be mixed and matched; i.e. key terms that don't need spare titles removed won't need the `:remove_title` method to be sent, & won't have anything else deleted on accident. 

An alternative would be to call a content transformation direction from outside of the title bake, as in:

```ruby
BakeGenericEocSection.v1(klass: 'key-terms', other args...)
BakeKeyTerms.v1(klass: 'key-terms')
RemoveTitle.v1(klass: 'key-terms')
OtherContentTransform.v1(klass: 'key-terms')
```

wherein BakeGenericEocSection just handles the titling, and each EocContentTransform method is its own direction that operates on just the section content. This would have the advantage of being more similar to how the MoveExercisesToEoc & BakeNumberedExercise directions already work. Furthermore, it prevents EocContentTransform from becoming a cluttered direction. The disadvantage of this implementation would be that it isn't quite as lightweight, and that the search for the sections that need transforming may need to be performed several times, which may slow down the bake.

lib/kitchen/directions/bake_generic_eoc_section/v1.rb
LostMojo: Hm, while most EOC content uses`<section>` elements, my book has at least two that use something else (Glossary and Solutions).  Would it be alright to have a slightly broader strategy?

LostMojo (commented):
I think the strategies might more applicable if they targeted the whole content, rather than dividing it up into sections.

lib/kitchen/directions/bake_generic_eoc_section/v1.rb
kswanson33: sure! I can adjust this so that any element can be passed in

kswanson33:
from a call with JP: This will be implemented so that `BakeEocCompositePage` (renamed) yields the sections that get moved to it. `EocContentTransform` methods will become their own directions. From the recipe, this can then be called with:

```ruby
BakeEocCompositePage.v1(args...) do |section|
  RemoveTitle.v1(section)
  KeyTerms.v1(section)
  OtherContentTransform.v1(section)
  ...
end
```

Common/existing directions, i.e. `BakeGlossary`, can be refactored to wrap these calls in Kitchen; that way, the refactor can be done without breaking changes, and our recipes don't have to get super complicated, especially since many books have common EOC sections, such as the glossary/key terms. 

Specs for existing eoc page directions should still pass. But new content transform directions can be more lightweight if desired: the EOC title shouldn't need to be tested specifically, and the content transformation alone can be targeted.

LostMojo (approved):

lib/kitchen/directions/move_eoc_content_to_composite_page/v1.rb
```diff
@@ -0,0 +1,19 @@
+# frozen_string_literal: true
+
+module Kitchen::Directions::MoveEocContentToCompositePage
+  class V1
+    renderable
+
+    def bake(metadata_source:, klass:, content:, append_to: nil, uuid_prefix: '.')
```
kswanson33, line 7: append_to is never nil

LostMojo (approved):

lib/kitchen/directions/bake_generic_eoc_section/v1.rb
```diff
@@ -0,0 +1,29 @@
+# frozen_string_literal: true
+
+module Kitchen::Directions::BakeGenericEocSection
```
jpslav, line 3: What might also help re the "what does `klass` mean?" comment above... should the name of this direction be different?  

The current name sounds like it is making a new section at the end of a chapter.  But I don't think that's it.  More that it is moving sections to a new container at the end of a chapter.  Would `MoveSectionsToEoc` be better? (or `MoveSectionsToNewEocContainer`) The direction would still yield sections allowing them to be baked.  Then maybe `klass` becomes `section_class` and then readers will be like "oh ok this variable is the class of the sections I want moved to the end".

(I see below that `klass` is used for multiple jobs which might make you not want to change it, but I think this one field being used for 4 different purposes only one of which is as a CSS class is probably a problem)

lib/kitchen/directions/bake_generic_eoc_section/v1.rb
```diff
@@ -0,0 +1,29 @@
...
+  class V1
+    renderable
+
+    def bake(chapter:, metadata_source:, klass:, append_to: nil, uuid_prefix: '.',
+             include_intro: true, &block)
+      section_clipboard = Kitchen::Clipboard.new
+      sections = if include_intro
+                   chapter.pages.search("$.#{klass}")
+                 else
+                   chapter.non_introduction_pages.search("$.#{klass}")
```
jpslav, line 13: IIRC using a `$` here doesn't do anything with a generic `search` call.  A place you would use a `$` is if you're trying to modify the selector for a specialized search call, like instead of `chapter.pages` you could say `chapter.pages("$.#{klass}")` which is the same as `chapter.search("div[data-type='page'].#{klass}")`.

Relatedly, this search as is will be without a tag name so it will be slower.  Should we change `klass` to `section_selector` and have callers pass in e.g. `'section.key-equations'`?  Such a name change in conjuction with renaming the direction would do a lot to clear up the purpose of the direction and what its variables mean... `MoveSectionsToEoc.v1(... section_selector: 'section.key-equations' ...)` makes it pretty obvious that you are moving `section.key-equations` elements to the end of the chapter.

lib/kitchen/directions/bake_generic_eoc_section/v1.rb
```diff
@@ -0,0 +1,29 @@
+# frozen_string_literal: true
+
+module Kitchen::Directions::BakeGenericEocSection
+  class V1
+    renderable
+
+    def bake(chapter:, metadata_source:, klass:, append_to: nil, uuid_prefix: '.',
+             include_intro: true, &block)
+      section_clipboard = Kitchen::Clipboard.new
+      sections = if include_intro
```
jpslav, line 10: Might make this `include_intro_page` for added clarity

lib/kitchen/directions/bake_generic_eoc_section/v1.rb
```diff
@@ -0,0 +1,29 @@
...
+                   chapter.pages.search("$.#{klass}")
+                 else
+                   chapter.non_introduction_pages.search("$.#{klass}")
+                 end
+      sections.each(&block) # transforms each section according to the given block
+      sections.cut(to: section_clipboard)
```
jpslav, line 16: `[enumerator].each` and `[enumerator].cut` is probably running the search twice.  It might be a little more efficient to cut the sections and then iterate on them inside the clipboard ([which is itself enumerable](https://github.com/openstax/kitchen/blob/main/lib/kitchen/clipboard.rb#L8)):

```ruby
pages = include_intro_page ? chapter.pages : chapter.non_introduction_pages
section_clipboard = pages.search(section_selector).cut
section_clipboard.each(&block)
```

A benefit of using these 3 lines to replace lines 9 through 16 is that you don't have to repeat the `search(section_selector)` call.

lib/kitchen/directions/eoc_composite_page_container/v1.rb
```diff
@@ -0,0 +1,19 @@
...
+    def bake(metadata_source:, klass:, content:, append_to:, uuid_prefix: '.')
+      @klass = klass
+      @metadata = metadata_source.children_to_keep.copy
+      @title = I18n.t(:"eoc.#{klass}")
+      @uuid_prefix = uuid_prefix
+      @content = content
+      @in_composite_chapter = append_to.is?(:composite_chapter)
+
+      append_to.append(child: render(file:
+        '../../templates/eoc_section_title_template.xhtml.erb'))
```
jpslav, line 16: Should we take `title` out of this name of this file?  This isn't the template for an end-of-chapter section title, it is the template for an end-of-chapter section.

lib/kitchen/directions/eoc_composite_page_container/v1.rb
```diff
@@ -0,0 +1,19 @@
+# frozen_string_literal: true
+
+module Kitchen::Directions::EocCompositePageContainer
+  class V1
+    renderable
+
+    def bake(metadata_source:, klass:, content:, append_to:, uuid_prefix: '.')
```
jpslav, line 7: `klass` here is not used as a CSS class -- it is used in an I18n key, as part of another klass, and in the UUID key.  That all of these uses need the same value is more of a coincidence and not a reason to actually make them the same variable.  So I think we should give it a different name, and I realize this means splitting it off from the `section_selector` variable up above.  I think splitting this one variable (that has many uses) into more than one variable is a good thing.

```ruby
def bake(metadata_source:, title_key:, uuid_key:, container_class_type:, content:, append_to:)
```

Note that by specifying the `uuid_key` you can get rid of the `uuid_prefix`.  I realize this makes the calls a little repetitive, but fortunately they are hidden inside other higher-level directions like `BakeChapterSummary` and won't be seen in the recipes themselves, generally.  Splitting out these variables will make the documentation and the readability better.

lib/kitchen/templates/eoc_section_title_template.xhtml.erb
```diff
@@ -3,7 +3,7 @@
     <span class="os-text"><%= @title %></span>
   </<%= @in_composite_chapter ? 'h3' : 'h2' %>>
```
jpslav, line 4: This can be DRYed up a hair

```ruby
<%- main_title_tag = @in_composite_chapter ? 'h3' : 'h2' 
-%>
<div class="os-eoc os-<%=@klass%>-container" data-type="composite-page" data-uuid-key="<%=@uuid_prefix%><%=@klass%>">
  <<%= main_title_tag %> data-type="<%= @in_composite_chapter ? 'title' : 'document-title' %>">
    <span class="os-text"><%= @title %></span>
  </<%= main_title_tag %>>

File: spec/directions/bake_chapter_key_equations_spec.rb [hunk 10]
```diff
@@ -40,11 +45,14 @@
...
     it 'works' do
+      described_class.v1(chapter: chapter, metadata_source: metadata_element)
       expect(
-        described_class.v1(chapter: chapter, metadata_source: metadata_element)
+        chapter
       ).to match_normalized_html(
         <<~HTML
           <div data-type="chapter">
+            <div data-type="page">
+          </div>
```
jpslav, line 55: Intentional to have page closing tag on next line unindented? Reads weirdly.

jpslav (changes requested):

CHANGELOG.md
```diff
@@ -6,6 +6,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+* Add a `BakeGenericEocSection` to allow for custom sections (minor)
+* Add a `EocCompositePageContainer` direction to handle creation of eoc page wrapper (minor)
+* Refactor `BakeChapterKeyEquations`, `BakeChapterGlossary`, `BakeChapterSummary`, `BakeFurtherResearch` (major)
```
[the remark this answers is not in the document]
kswanson33: only in that the translation keys are being standardize to use the `"eoc.#{classname}"` format, but locale files will have to be updated for dependent recipes

lib/kitchen/directions/bake_generic_eoc_section/v1.rb
kswanson33: That makes sense -- by 4 different purposes, do you mean the title lookup & the middle part of the wrapper class? I can change the variable name (`klass` -> `section_class`) here and in the other eoc sections.

As for the direction name for `BakeGenericEocSection`, what do you think about `MoveCustomSectionToEocContainer` to emphasize that it differs from other more specific directions in that it allows the dev to pass in more custom parameters/modify the yielded sections?

lib/kitchen/directions/bake_generic_eoc_section/v1.rb
kswanson33: Interesting ... is the `$` just a concatenation character then?

I have previously experimented with including the `section` element name in the selector. Here's what I've tried:

1. Originally, I had it pass in `klass` and then search with `.search('section.klass')`. This changed because @LostMojo mentioned some books have `div`s instead of `section`s. 
2. Next, I tried a call to pass in `selector` instead of `klass`, intending for something like `section.klass` to be given. But this would need some string parsing, because, as you pointed out above, the `klass` is used more than just as a CSS selector for the sections -- it's also used as a title lookup key and as part of the classes on the wrapper. So I would have to break the `klass` portion out of `section.klass` at some point. This is definitely doable, but since a `selector` parameter doesn't necessarily indicate an exclusively `name.class` format, I thought it would be better to specify what's expected. 

I can switch the implementation to 2) though, since I agree it would be faster. The name `section_selector` may make it more clear that the selector has an expected format. What do you think?

lib/kitchen/directions/bake_generic_eoc_section/v1.rb
kswanson33: There seems to be some kind of bug that's affecting this implementation: when I call `section.ancestor(:page).count_in(:chapter)` from the `block`, if the `section` has been cut to a clipboard, then the chapter count always returns the same number (i.e. if there are 3 pages in the chapter it always returns `3`, if there are 2 it always returns `2`, etc). Digging into this a little, I've discovered that `section.ancestor(:page)` does seem to return the correct page, so I've been stumped so far on why this page seems to always be getting counted as the last page of the chapter.

This affects the section content transform for Further Research, since those sections get a header link added that contains the `chapter.page` count of the section's origin page. I think a couple of other eoc sections (that are yet to be refactored) also do things with the section's origin page. Until we figure out this bug, I think the block has to be passed in before the sections get cut to the clipboard.

lib/kitchen/directions/eoc_composite_page_container/v1.rb
kswanson33: 👀 for clarification -- an example of this being called on, say, Further Research would be:

`bake(metadata_source: a MetadataElement, title_key: 'further-research', uuid_key: 'further-research'/'.further-research', container_class_type: 'section'/'div'/etc, content: '...', append_to: KitchenElement)`

I think a better name for `container_class_type` would be `section_element_name`, if I'm interpreting the purpose of that field correctly.

spec/directions/bake_chapter_key_equations_spec.rb
kswanson33: the spec fails unless it's formatted this way unfortunately

lib/kitchen/directions/bake_eoc_section_content/remove_section_title.rb
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+
+module Kitchen
+  module Directions
+    module RemoveSectionTitle
+      def self.v1(section:)
+        section.first('h3').trash
```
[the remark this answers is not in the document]
kswanson33: It has data-type="title" in all the books I've looked at, but the spec for Key Equations doesn't include that data type. Not sure if that means there's a book that's an exception? I agree that this seems like a better selector, but there's a chance it could be a breaking change.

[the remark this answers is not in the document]
jpslav: Ok I yield to you on this

lib/kitchen/directions/eoc_composite_page_container/v1.rb
jpslav: There's a funky `/` in your example?

`container_class_type` is what gets interpolated into `os-blah-container`, so I don't know if `section_element_name` is right?  An element name is something like `h3`.

lib/kitchen/directions/bake_generic_eoc_section/v1.rb
jpslav: `Custom` is fine.

lib/kitchen/directions/bake_generic_eoc_section/v1.rb
jpslav: Yes on (2) and as you saw below recommending we don't reuse this field for the other purposes, so string parsing won't be necessary

jpslav: The `$` character, when passed to a specific enumerator (like `pages`, not a general enumerator like `search`) gets replaced with the CSS for that specific enumerator, in the case of pages it gets replaced with `div[data-type="page"]`

lib/kitchen/directions/bake_generic_eoc_section/v1.rb
jpslav: ok, if don't fix now, you may make a PR with a failing spec so we can get to it later.

lib/kitchen/directions/eoc_composite_page_container/v1.rb
kswanson33: oh the slashes are pseudocode

kswanson33: I believe the above isn't a full set of the information this method would need. For Further Research, an example would be:

- title_key: 'further-research'
- uuid_key: either 'further-research' or '.further-research'
- container_class_type: 'os-further-research-container'
- section_selector: 'section.further-research' or 'div.further-research'
- metadata_source
- content
- append_to

Just to double-check, are you advocating that the first 4 items in that list should all be separate fields, despite the repetition?

lib/kitchen/directions/bake_eoc_section_content/remove_section_title.rb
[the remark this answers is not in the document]
kswanson33: haha I did change it to data-type="title" ... 🤞

kswanson33:
To clarify the above: I believe the only two pending comments are regarding [the fields for eoc composite page container](https://github.com/openstax/kitchen/pull/280#discussion_r664788989), & [the implementation for these lines, specifically re: switching their order](https://github.com/openstax/kitchen/pull/280/files#diff-55bb99aa6cfe598b9a7821881c8519a8b887b1f23863aba35706b5a7cb7d2523R14-R15). 

The latter is currently held up by a bug, & I will PR to show failing specs after this PR is merged.

lib/kitchen/directions/eoc_composite_page_container/v1.rb
kswanson33: OOPS WAIT. I see that we don't need `section_selector` in `EocCompositePageContainer`. It would be the newly-renamed `MoveCustomSectionToEocContainer` that would need all of the first 4 fields passed in. Again, to double-check, is that your preferred implementation?

jpslav (approved):

PR #281, opened 2021-06-17
Optimize selectors 

KarinaMendez2:


lib/kitchen/directions/move_solutions_to_answer_key/strategies/default.rb
```diff
@@ -11,7 +11,7 @@ def bake(chapter:, append_to:)
 
       def bake_section(chapter:, append_to:)
         @classes.each do |klass|
-          chapter.search(".#{klass} [data-type='solution']").each do |solution|
+          chapter.search("#{klass} div[data-type='solution']").each do |solution|
```
jpslav, line 14: if now passing in `div.some_class` then need to change `@classes` to `@selectors` and `klass` to `selector` b/c these aren't just classes any more.

jpslav (changes requested):

jpslav (approved):

brittany-johnson (approved):

PR #283, opened 2021-06-18
add cache_methods

jpslav:
You know how we are manually caching things in recipes with lines like `metadata = book.metadata`?  This PR gives a way to do that for more methods more easily.  E.g. you can say

```ruby
Kitchen::BookElement.cache_methods(:metadata, :chapters)
Kitchen::PageElement.cache_methods(:pages, :non_introduction_pages)
```

at the top of your recipe and this PR will cache the results from calls like `book.chapters` and then for each chapter the subsequent calls to `chapter.pages`.  This caching is only recommended for search results that don't change (obviously) -- i.e. chapters don't really move around in our baking, so it _should_ be ok to use this.  Definitely check the results and check the profiling time changes!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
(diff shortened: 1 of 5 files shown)


jpslav:
This isn't going to work as I wanted.

PR #289, opened 2021-06-21
Fix for baking nested notes

KarinaMendez2:


lib/kitchen/directions/bake_notes/bake_autotitled_notes.rb
```diff
@@ -8,6 +8,14 @@ def self.v1(book:, classes:)
           next unless (note.classes & classes).any?
 
           bake_note(note: note)
+
+          next unless note.notes.present?
```
jpslav, line 12: is this needed?  if there are no `note.notes` then the following loop will do nothing anyway

lib/kitchen/directions/bake_notes/bake_autotitled_notes.rb
```diff
@@ -8,6 +8,14 @@ def self.v1(book:, classes:)
...
+
+          note.notes.each do |nested_notes|
```
jpslav, line 14: doesn't `book.notes` above already find nested notes and bake them?

I actually checked out this branch, commented out the changes, and the test still passed so I don't think this PR is necessary? :-)

jpslav (changes requested):

lib/kitchen/directions/bake_notes/bake_autotitled_notes.rb
KarinaMendez2: it wasn't baking the nested notes in calculus vol 3 until I did this PR

jpslav (approved):

PR #290, opened 2021-06-21
Patch for array to add prefix

KarinaMendez2:


lib/kitchen/patches/array.rb
```diff
@@ -0,0 +1,18 @@
...
+  # Receives a string to add as prefix in each item of the array
+  #
+  # @return [String]
+  #
+  def prefix(string)
+    result = []
+
+    each { |item| result << "#{string}#{item}" }
+
+    result
```
jpslav, line 16: Instead of starting a new array and pushing onto it using `each`, use `map`:

```ruby
map { |item| "#{string}#{item}" }
```

lib/kitchen/patches/array.rb
```diff
@@ -0,0 +1,18 @@
+# frozen_string_literal: true
+
+# Monkey patches for +Array+
+#
+class Array
+
+  # Receives a string to add as prefix in each item of the array
+  #
+  # @return [String]
```
jpslav, line 9: `@return [Array<String>]`

lib/kitchen/patches/array.rb
```diff
@@ -0,0 +1,18 @@
+# frozen_string_literal: true
+
+# Monkey patches for +Array+
+#
+class Array
+
+  # Receives a string to add as prefix in each item of the array
```
jpslav, line 7: Make it clear that this is returning a new array (not modifying this array in place)

CHANGELOG.md
```diff
@@ -6,8 +6,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 * Changed selector expected by `default strategy` in `move_solutions_to_answer_key` to optimize search (major)
-
 * Fix Rubocop GitHub Action's regular expression used to select files to lint (patch)
+* Added monkey patch for array to be able to add a prefix: `%w[multiple-choice true-false].prefix('section.')` (patch)
```
jpslav, line 10: Adding a feature is minor.  Bug fixes are patches.

jpslav (changes requested):

jpslav (approved):

PR #301, opened 2021-07-05
implementing V2 inside V1 instead

jpslav:


wilkus27:
@jpslav I made some fixes and cleaned up code. It's ready for review :slightly_smiling_face: Sorry it took so long, but I was working on other issues with higher priority and couldn't look at this earlier.

lib/kitchen/directions/bake_index/main.rb
```diff
@@ -3,12 +3,8 @@
 module Kitchen
   module Directions
     module BakeIndex
-      def self.v1(book:)
-        V1.new.bake(book: book)
-      end
-
-      def self.v2(book:, types:)
-        V2.new.bake(book: book, types: types)
+      def self.v1(book:, types:, uuid_prefix: nil)
```
kswanson33, line 6: I believe the default value to `types` should be set here as well/instead of in `def bake` in the v1 file.

lib/kitchen/directions/bake_index/v1.xhtml.erb
```diff
@@ -1,9 +1,9 @@
-<div class="os-eob os-index-container " data-type="composite-page" data-uuid-key="index">
+<div class="os-eob <%= @container_class %> " data-type="composite-page" data-uuid-key="<%= @uuid_key %>">
```
kswanson33, line 1: the space after the interpolated `container_class` tag seems to be creating trailing whitespace in the classes; might be a good idea to clean that up

CHANGELOG.md
```diff
@@ -6,7 +6,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
-* Create V2 for `BakeIndex` to bake different types of indexes (minor)
+* Modify `BakeIndex` to bake multiple types of indexes (major)
```
kswanson33, line 9: is this major? it seems backwards-compatible to me

kswanson33 (changes requested):

CHANGELOG.md
wilkus27: Hmm :thinking: There should be an error that `types` key is missing.
I tried to bake Calculus and I got this:

image

kswanson33: I believe that error should not be happening now that a default value has been added to `BakeIndex#v1`

kswanson33 (approved):
This should possibly still be changed to minor ([see comment](https://github.com/openstax/kitchen/pull/301#discussion_r684290615)) but I leave that up to you! The rest looks good

CHANGELOG.md
wilkus27: You're right :wink: There is no error now. I changed log to `minor`.

PR #303, opened 2021-07-07
Refactor remaining eoc sections to use new method

kswanson33:


lib/kitchen/directions/move_exercises_to_eoc/v3.rb
```diff
@@ -3,48 +3,23 @@
...
-        sections.each do |exercise_section|
-          exercise_section.first("[data-type='title']")&.trash
-
-          section_title = Kitchen::Directions::EocSectionTitleLinkSnippet.v1(page: page)
-
-          exercise_section.exercises.each do |exercise|
-            exercise.document.pantry(name: :link_text).store(
-              "#{I18n.t(:exercise_label)} #{chapter.count_in(:book)}.#{exercise.count_in(:chapter)}",
-              label: exercise.id
-            )
```
kswanson33, line 30: I removed this because exercises are already stored in the pantry in `BakeNumberedExercises`

lib/locales/en.yml
```diff
@@ -15,13 +15,12 @@ en:
   chapter_review: Chapter Review
   learning_objectives: Learning Objectives
   stepwise_step_label: Step
-  references: References
+  references: References # is this used somewhere outside eoc?
```
kswanson33, line 18: thoughts? 😄

spec/directions/bake_chapter_key_concepts_spec.rb
```diff
@@ -16,7 +18,7 @@
         <div data-type="page">
           <h1 data-type="document-title" id="page1TitleId">Page 1</h1>
           <section id="sectionId1" class="key-concepts">
-            <h3 data-type="document-title">WWF History</h3>
+            <h3 data-type="title">WWF History</h3>
```
kswanson33, line 21: I switched to using the `RemoveSectionTitle` direction to delete this element, which searches for the first `data-type="title"` element instead of an `h3`. All books I looked at, including the key concepts section in uphysics and calculus, used `data-type="title"`, so I'm hoping the `data-type="document-title"` attribute in these specs is just a typo. But please let me know if you've seen anything different.

lib/kitchen/directions/bake_chapter_summary.rb
```diff
@@ -15,7 +15,6 @@ def self.v1(chapter:, metadata_source:, klass: 'summary', uuid_prefix: '.')
       end
 
       class V1
-        renderable
         def bake(chapter:, metadata_source:, uuid_prefix: '.', klass: 'summary')
           summaries = Clipboard.new
```
wilkus27, line 20: I baked pl-psychology on this branch and I noticed additional attribute in H3 headers from Summary:

image

Probably it's related with previous changes, not specifically with this PR :thinking:

lib/locales/en.yml
wilkus27: Yes, in American Government (I guess :thinking:) and in Polish Psychology section References is in EOB. We use there `BakeReferences` module.

lib/locales/pl.yml
```diff
@@ -19,9 +19,10 @@ pl:
   eoc_solutions_title: Rozwiązania
   eob_index_title: Skorowidz rzeczowy
   eob_index_symbols_group: Symbole
-  references: Bibliografia
+  references: Bibliografia # is this used somewhere outside eoc?
   eoc:
     glossary: Kluczowe pojęcia
     key-equations: Najważniejsze wzory
     summary: Podsumowanie
+    references: Bibliografia
```
wilkus27, line 27: As I write in my previous comment in pl-psychology References is not EOC section, so we can remove this :slightly_smiling_face:

wilkus27 (changes requested):
Looks good :relaxed: I added a few comments related with Polish Psychology.

lib/locales/pl.yml
kswanson33: I believe that `eoc.references` is used by `BakeChapterReferences`, which I refactored in this PR. The `EocCompositePageContainer` direction requires that the title use the translation key `eoc.{container_key}` where `container_key` comes from the `os-{container_key}-container` wrapper class. So it seems that both `references` and `eoc.references` are necessary now.

So I'll take out the comment & leave everything else in, thank you for clarifying!

lib/kitchen/directions/bake_chapter_summary.rb
kswanson33: Good catch... 🤔 You're right, [this is where that change was introduced.](https://github.com/openstax/kitchen/pull/280/files#diff-ee054ac0aca7a439df006e3f489616ea44c8785f0e188b0577b8ac2f54642b44R28-R30) 

I manually added the `itemprop` attribute because it showed up in the test data; `EocTitleLinkSnippet` doesn't add it. Now I'm wondering if we should change the implementation of `EocTitleLinkSnippet`? It looks like, for the `BakeChapterSummary` direction, this title was being added with `page.title.copy`, instead of being manually constructed as it is in `EocTitleLinkSnippet`, which is probably letting it have the full set of attributes.

Should we tackle this in a separate PR?

wilkus27: Yes, I think it's a good idea to take care of this in a separate PR :+1: Probably we should investigate why section title link was added in a different way in Summary. If it was intentional or someone just came up with different idea.

wilkus27 (approved):

LostMojo (approved):

PR #307, opened 2021-07-14
Philsophy changes

brittany-johnson:


lib/kitchen/directions/bake_further_reading.rb
```diff
@@ -0,0 +1,20 @@
...
+
+    Kitchen::Directions::EocCompositePageContainer.v1(
+      container_key: klass,
+      uuid_key: "#{uuid_prefix}#{klass}",
+      metadata_source: metadata_source,
+      content: further_reading.paste,
+      append_to: chapter
+    )
+  end
+end
```
kswanson33, line 20: This seems like an opportunity to use `MoveCustomSectionToEocContainer`. This direction [can be passed an optional block argument](https://github.com/openstax/kitchen/blob/main/lib/kitchen/directions/move_custom_section_to_eoc_container/v1.rb#L12), which transforms each section before it's moved to the eoc container. This way, basic content transformations con be done from outside the direction. 

Not every eoc section is simple enough to use this -- if the content requires more elaborate transformation (i.e. a wrapper), the best we can do is `EocCompositePageContainer`. But in this case, it looks like this section just needs to remove a redundant title. [Here](https://github.com/openstax/recipes/pull/46/files#diff-b241fef3bc8dbd18e144e07149bfefcbb936d880a72c2ce0fb2487ff14b16c56R32-R41) is an example of a similar thing being done for poli sci. 

The advantage of this approach is that the section can be transformed entirely on the recipe side, and a new direction doesn't have to be made in Kitchen. This is good because it's less work for devs, we don't have to come up with test data, Kitchen doesn't get cluttered with one-off eoc sections, and the recipe won't be held back by needing a Kitchen release for a new eoc section. If a few other books have the same eoc section, though, then we might want to refactor the direction into Kitchen.

tl;dr we probably don't need to add `BakeFurtherReading` to Kitchen just yet if philo is the only book to use it

lib/locales/en.yml
```diff
@@ -22,6 +22,8 @@ en:
   eoc_composite_metadata_title: Chapter Review
   eoc_solutions_title: Solutions
   eoc_suggested_reading: Suggestions for Further Study
+  eoc_further_research_title: Further Research
+  eoc_further_reading: Further Reading
```
kswanson33, line 26: Like notes, eoc sections are now being titled with `eoc.class-name` syntax. Looks like you're already doing that on [these lines](https://github.com/openstax/kitchen/pull/307/files#diff-887c311739256a99225ab8153707412184ba7b8b571ad90d45462344f25ab86cR9-R11) in the spec, but the locale should match. Also, if these titles are just used in philosophy, they might belong in the philosophy-specific locale file.

lib/kitchen/directions/bake_chapter_references/v1.rb
```diff
@@ -2,11 +2,11 @@
 
 module Kitchen::Directions::BakeChapterReferences
   class V1
-    def bake(chapter:, metadata_source:, uuid_prefix: '.', klass: 'references')
-      bake_page_references(page: chapter.introduction_page)
+    def bake(chapter:, metadata_source:, uuid_prefix: '.', klass: 'references', module_title: true)
+      bake_page_references(page: chapter.introduction_page, module_title: module_title)
```
kswanson33, line 6: see other comment, but could `MoveCustomSectionToEocContainer` also be used here instead of `BakeChapterReferences`? it seems like with `module_title=false`, all this does to the content of the section is just removing the redundant title. not sure I'm reading this right

kswanson33 (changes requested):
This works, but I think we can get away with using less code to do it 👀

brittany-johnson:
No longer need this PR after replacing directions w/ `MoveCustomSectionToEocContainer`

PR #314, opened 2021-07-19
BakeNumberedTablesV1 tweaks

LostMojo:
Tweaks to
- include non-span titles
- remove long blank spaces in captions

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+* Tweak `BakeNumberedTables.v1` to cut extra white space and include non-span titles (major)
 * Update `BakeFirstElements` to also add the `.has-first-inline-list` class (minor)
 
 ## [9.1.0] - 2021-07-16
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
-* Tweak `BakeNumberedTables.v1` to cut extra white space and include non-span titles (major)
+* Tweak `BakeExample` to not touch unbaked titles in captions (major)
+* Tweak `BakeNumberedTables.v1` to cut extra white space (minor)
 * Update `BakeFirstElements` to also add the `.has-first-inline-list` class (minor)
 
 ## [9.1.0] - 2021-07-16
--- a/lib/kitchen/directions/bake_example.rb
+++ b/lib/kitchen/directions/bake_example.rb
@@ -22,7 +22,8 @@ module Kitchen
 
         example.titles.each do |title|
           if title.parent.has_class?('os-caption-container') || \
-             title.parent.has_class?('os-caption')
+             title.parent.has_class?('os-caption') || \
+             title.parent.name == 'caption'
             next
           end
 
--- a/lib/kitchen/directions/bake_numbered_table/v1.rb
+++ b/lib/kitchen/directions/bake_numbered_table/v1.rb
@@ -17,7 +17,7 @@ module Kitchen::Directions::BakeNumberedTable
       end
 
       if (caption = table.caption&.cut)
-        caption.children = '' if caption.children.to_s.blank?
+        caption.replace_children with: '' if caption.children&.to_s&.blank?
         new_caption = <<~HTML
           \n<span class="os-caption">#{caption.children}</span>
         HTML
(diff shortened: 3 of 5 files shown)


lib/kitchen/directions/bake_numbered_table/v1.rb
```diff
@@ -10,13 +10,14 @@ def bake(table:, number:, always_caption: false)
...
 
-      if (title = table.first("span[data-type='title']")&.cut)
+      if (title = table.first("[data-type='title']")&.cut)
         caption_title = <<~HTML
           \n<span class="os-title" data-type="title">#{title.children}</span>
         HTML
       end
 
       if (caption = table.caption&.cut)
+        caption.replace_children with: '' if caption.children&.to_s&.blank?
```
kswanson33, line 20: can you update the specs to cover the added logic?

kswanson33 (changes requested):

lib/kitchen/directions/bake_numbered_table/v1.rb
kswanson33: oh -- is this the same behavior as you'd get if you passed `true` to `always_caption`?

LostMojo: That is a _good point_, thank you. Yeah, that might end up with an accidental caption

LostMojo: I will fix that

lib/kitchen/directions/bake_numbered_table/v1.rb
```diff
@@ -3,20 +3,20 @@
...
-    def bake(table:, number:, always_caption: false)
+    def bake(table:, number:, always_caption: false, title_css: "span[data-type='title']")
       Kitchen::Directions::BakeTableBody.v1(table: table, number: number)
 
       # TODO: extra spaces added here to match legacy implementation, but probably not meaningful?
       new_caption = ''
       caption_title = ''
 
-      if (title = table.first("span[data-type='title']")&.cut)
+      if (title = table.first(title_css)&.cut)
```
kswanson33, line 13: I really like that you've thought of a way to keep the selector specific (& therefore fast!). What do you think of instead defining a `TableElement#title` or `TableElement#caption_title` method? This would allow the selector to be updated/added to easily, for example, to `'span[data-type="title"], h4[data-type="title"]'` (I think that's `or`-syntax for a css selector, but I'd have to double-check). I believe `CompositePageElement#title` does something similar.

lib/kitchen/directions/bake_numbered_table/v1.rb
```diff
@@ -3,20 +3,20 @@
...
         caption_title = <<~HTML
           \n<span class="os-title" data-type="title">#{title.children}</span>
         HTML
       end
 
-      if (caption = table.caption&.cut)
+      if (caption = table.caption&.cut) && !caption&.children&.to_s&.blank?
```
kswanson33, line 19: good catch! thank you for covering this in specs as well!

kswanson33 (changes requested):

lib/kitchen/directions/bake_numbered_table/v1.rb
LostMojo: Actually, I think I got the root cause of this wrong-- the assembled uses all span titles.... but there's something along the way that turns some of them into h4s.

lib/kitchen/directions/bake_example.rb
```diff
@@ -22,7 +22,8 @@ def self.v1(example:, number:, title_tag:, numbered_solutions: false)
 
         example.titles.each do |title|
           if title.parent.has_class?('os-caption-container') || \
-             title.parent.has_class?('os-caption')
+             title.parent.has_class?('os-caption') || \
+             title.parent.name == 'caption'
```
kswanson33, line 26: miiiight be good to enforce this with a spec in case someone comes across it later and tries to change it -- i think the `BakeExample` spec has something in there already [about tables](https://github.com/openstax/kitchen/blob/main/spec/directions/bake_example_spec.rb#L136), so this would just be a matter of adding a different table input format

CHANGELOG.md
```diff
@@ -6,6 +6,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+* Tweak `BakeExample` to not touch unbaked titles in captions (major)
```
kswanson33, line 9: this should be minor (or patch?) since, although it is changing the implementation of `BakeExample`, it's just a bug fix & not something that will like. impact or break other books

kswanson33 (changes requested):

lib/kitchen/directions/bake_example.rb
LostMojo: done and done

kswanson33 (approved):

PR #317, opened 2021-07-20
Move custom sections to EOB

KarinaMendez2:


lib/kitchen/directions/bake_references/v3.rb
```diff
@@ -0,0 +1,26 @@
...
+                <span class="os-divider"> </span>
+                <span class="os-text" data-type="" itemprop="">#{page.title_text} </span>
+              HTML
+            )
+          end
+        end
+      end
+    end
+  end
+end
```
wilkus27, line 26: I'm wondering if we can create something like that https://github.com/openstax/kitchen/blob/b36c33685e89c8b797d6a6fee6dbf1953725d43d/lib/kitchen/directions/eoc_section_title_link_snippet.rb for this code :thinking: @kswanson33 you were working on that. What do you think?

lib/kitchen/directions/eob_composite_page_container/v1.rb
```diff
@@ -0,0 +1,19 @@
...
+  class V1
+    renderable
+
+    def bake(container_key:, uuid_key:, metadata_source:, content:, append_to:)
+      @title = I18n.t(:"eoc.#{container_key}")
+      @uuid_key = uuid_key
+      @container_class_type = container_key
+      @metadata = metadata_source.children_to_keep.copy
+      @content = content
+      @in_composite_chapter = append_to.is?(:composite_chapter)
```
wilkus27, line 13: I'm not sure if we need this here :thinking: In EOB only composite chapter is Answer Key. The question is whether we want to use this module there? I didn't work on any Answer Key in Kitchen, so it's hard for me to have any opinion about that, so maybe someone else have any thoughts about advantages and disadvantages of using that in solutions.

lib/kitchen/templates/eob_section_template.xhtml.erb
```diff
@@ -0,0 +1,11 @@
...
+<div class="os-eob os-<%=@container_class_type%>-container" data-type="composite-page" data-uuid-key="<%= @uuid_key %>">
+  <<%= main_title_tag %> data-type="<%= @in_composite_chapter ? 'title' : 'document-title' %>">
+    <span class="os-text"><%= @title %></span>
+  </<%= main_title_tag %>>
+  <div data-type="metadata" style="display: none;">
+    <h1 data-type="document-title" itemprop="name"><%= @title %></h1>
+    <%= @metadata.paste %>
+  </div>
+  <%= @content %>
+</div>
```
wilkus27, line 11: Actually there is already such template: https://github.com/openstax/kitchen/blob/b8aadf6e0bdff60f0b4fc3ee37fa23f81203c443/lib/kitchen/templates/eob_section_title_template.xhtml.erb . Maybe title is a little bit misleading.

First line from your template I think is not necessary in EOB. As I write in my previous comment, the only one composite chapter is Answer Key. Even we decide to use this for Answer Key composite pages I think they also have `h2` headers, not `h3`.

spec/directions/eob_composite_page_container/main_spec.rb
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+
+require 'spec_helper'
+
+RSpec.describe Kitchen::Directions::EocCompositePageContainer do
+  it 'calls v1' do
+    expect_any_instance_of(Kitchen::Directions::EocCompositePageContainer::V1).to receive(:bake)
```
wilkus27, line 7: There is a  typo here. It should be `Eob` in `EocCompositePageContainer`

spec/directions/eob_composite_page_container/v1_spec.rb
```diff
@@ -0,0 +1,75 @@
+# frozen-string-literal: true
+
+require 'spec_helper'
+
+RSpec.describe Kitchen::Directions::EocCompositePageContainer do
```
wilkus27, line 5: Here also should be `Eob` :wink:

wilkus27 (changes requested):
@KarinaMendez2 I added a few thoughts.

lib/kitchen/directions/bake_references/v3.rb
kswanson33: agreed! it looks like it could be the same thing -- does it work to substitute in `EocSectionTitleLinkSnippet` here?

lib/kitchen/directions/eob_composite_page_container/v1.rb
```diff
@@ -0,0 +1,19 @@
+# frozen_string_literal: true
+
+module Kitchen::Directions::EobCompositePageContainer
```
kswanson33, line 3: I am wondering if this implementation is the same as EocCompositePageContainer. If so, do we need a separate EOB class, or could we just use EOC? It would be good to reuse as much code as possible

lib/kitchen/directions/move_custom_section_to_eob_container/main.rb
```diff
@@ -0,0 +1,37 @@
+# frozen_string_literal: true
+
+# rubocop:disable Metrics/ParameterLists
+# More parameters are ok here because these generic classes DRY up a lot of other code
+module Kitchen
+  module Directions
+    module MoveCustomSectionToEobContainer
```
kswanson33, line 7: same as above -- could we also use EOC for this?

kswanson33 (changes requested):
I agree that it would be good to refactor EOB stuff to be similar to how EOC works! As part of my investigation into EOC sections, I found that it was important to also refactor the sections that already existed in Kitchen, because that gave me some guidance about what features the new directions needed in order to be backwards compatible. How do you feel about doing something similar for EOB? There are fewer EOB sections I believe, so hopefully it won't be too involved. 

I notice that your approach so far is very similar to EOC. If it turns out that all of the EOC directions also work for EOB pages, then that's great, and we should use those directions instead of creating new ones for EOB. If EOB pages end up having slightly different needs, then we should create new directions, and refactor existing EOB directions accordingly.

Let me know what you find out about which approach seems best as you continue to investigate this!

lib/kitchen/directions/bake_references/v3.rb
KarinaMendez2: I actually don't use links in these titles, so I don't think I can use that snippet without modifying it

lib/kitchen/directions/bake_references/v3.rb
```diff
@@ -0,0 +1,37 @@
...
+            title.name = 'h2'
+            title['data-type'] = 'document-title'
+            title['id'] = page.title.copied_id
+            title.replace_children(with:
+              <<~HTML.chomp
+                <span class="os-number">#{page.title.children[0].text}</span>
+                <span class="os-divider"> </span>
+                <span class="os-text" data-type="" itemprop="">#{page.title_text} </span>
+              HTML
+            )
```
kswanson33, line 26: Did you see that there's a [v2](https://github.com/openstax/kitchen/blob/main/lib/kitchen/directions/eoc_section_title_link_snippet.rb#L19) that might help here? Looks like the only difference is `h2` or `h3` in the title element, but that could be modified after the fact.

kswanson33 (changes requested):
This approach works, but ultimately EOB sections should be refactored similarly to how EOC sections have been.

lib/kitchen/directions/eoc_section_title_link_snippet.rb
```diff
@@ -4,30 +4,31 @@ module Kitchen
...
+      def self.title_snipet(page:, title_tag: 'h3')
+        chapter = page.ancestor(:chapter)
+        <<~HTML
+          <#{title_tag} data-type="document-title" id="#{page.title.copied_id}">
+            <span class="os-number">#{chapter.count_in(:book)}.#{page.count_in(:chapter)}</span>
+            <span class="os-divider"> </span>
+            <span class="os-text" data-type="" itemprop="">#{page.title_text}</span>
+          </#{title_tag}>
+        HTML
+      end
```
kswanson33, line 31: we discussed this & i think this could be 1 function with a couple of different specification options!

lib/kitchen/directions/bake_chapter_introductions.rb
```diff
@@ -21,12 +21,14 @@ def self.v1(book:, bake_chapter_objectives: true)
...
-              </div>
-            HTML
+            if bake_chapter_outline
+              chapter_outline_html = <<~HTML
+                <div class="os-chapter-outline">
+                  <h3 class="os-title">#{I18n.t(:chapter_outline)}</h3>
+                  #{outline_items_html}
+                </div>
+              HTML
+            end
```
wilkus27, line 31: Is this something from another PR? :thinking:

spec/directions/bake_references/v3_spec.rb
```diff
@@ -0,0 +1,85 @@
...
+        <div data-type="page">
+          #{metadata_element}
+          <div data-type="document-title" id="someId1">1.1 Page</div>
+          <p>
+            <a href="#auto_12345" data-type="cite">
+              <div data-type="note" class="reference" display="inline" id="auto_12345">
+                <h3 data-type="title">Reference 1</h3>
+              </div>
+            </a>
+            <a href="#auto_54321" data-type="cite">
```
wilkus27, line 26: I'm not sure here :thinking: Is V3 is created for Anatomy? If yes, there are no notes there. We are moving sections from modules to EOB:

image

wilkus27 (approved):
Looks good :wink:

lib/kitchen/directions/bake_references/main.rb
```diff
@@ -6,17 +6,49 @@ module Directions
...
+
+        section(
           book: book,
-          metadata_source: metadata_source
+          metadata_source: metadata_source,
+          chapter_area_references: chapter_area_references.paste
+        )
+      end
+
+      def self.section(book:, metadata_source:, chapter_area_references:, plural: true)
```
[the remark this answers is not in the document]
kswanson33: actually, this method doesn't seem to do much but wrap the call to `CompositePageContainer`; personally, I don't think it's worth the few lines of code it saves, especially since it makes the implementation more confusing to read. 

Is there a reason why the call to `CompositePageContainer` can't be in `V1#bake` etc?

kswanson33 (changes requested):

spec/directions/bake_references/main_spec.rb
```diff
@@ -3,15 +3,69 @@
...
-      .with(book: 'book1', metadata_source: 'metadata')
-    described_class.v2(book: 'book1', metadata_source: 'metadata')
+      .with(book: book1)
+    described_class.v2(book: book1, metadata_source: metadata_element)
+  end
+
+  it 'calls v3' do
+    expect_any_instance_of(Kitchen::Directions::BakeReferences::V3).to receive(:bake)
+      .with(book: book1)
+    described_class.v3(book: book1, metadata_source: metadata_element)
```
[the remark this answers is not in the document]
KarinaMendez2: since each version has more code than just calling each class bake, it crashes when it gets to [this line](https://github.com/openstax/kitchen/blob/74f45e3d95955efa018239d93a49573ae61dbe59/lib/kitchen/directions/bake_references/main.rb#L11) if it's not real data

lib/kitchen/templates/composite_page_template.xhtml.erb
```diff
@@ -1,5 +1,5 @@
-<% main_title_tag = @in_composite_chapter ? 'h3' : 'h2' %>
-<div class="os-eoc os-<%=@container_class_type%>-container" data-type="composite-page" data-uuid-key="<%= @uuid_key %>">
+<% main_title_tag = @in_composite_chapter ? 'h3' : @is_eoc ? 'h2' : 'h1' %>
```
kswanson33, line 1: this is getting complicated -- can `main_title_tag` become a template variable, and the logic here be moved to the class?

spec/directions/composite_page_container/v1_spec.rb
```diff
@@ -2,7 +2,7 @@
 
 require 'spec_helper'
 
-RSpec.describe Kitchen::Directions::EocCompositePageContainer do
+RSpec.describe Kitchen::Directions::CompositePageContainer do
```
kswanson33, line 5: since it looks like a bit of logic has been added to this class in order to accommodate EOB stuff, could a test be added for that here? I know it's already implicitly tested by `BakeReferences`, but I've found it's good to test things directly to catch things that might not be caught by the implicit case. I wouldn't be as worried about it if it was straightforward, but all of the cases in the template for `CompositePageContainer`are starting to get little hairy 💇

kswanson33 (changes requested):
found 2 more things when I looked this over 🤦 thanks for your patience!

kswanson33 (approved):

Issue #284, opened 2021-06-21
Remove paragraphs from Key Terms


wilkus27:
In books created by Adaptarr descriptions of definitions from Key Terms contain paragraphs: 

![image](https://user-images.githubusercontent.com/40228854/122738855-a9209b80-d282-11eb-950b-16bf9678a2ea.png)

It breaks styles:

![image](https://user-images.githubusercontent.com/40228854/122738972-c81f2d80-d282-11eb-9655-7e2952e14b78.png)

Text should be put directly in `dd` tag, like it is done in books created by other tools.

Comment why this problem is not resolved in Adaptarr:

> After the discussion, we would like it to be solved in the recipes. We've designed the schema to be always the same and since there might be more paragraphs in the key terms, we think that we shouldn't change how it works for a single paragraph.

PR #327, opened 2021-07-27
Remove para from term description

wilkus27:
Issue: https://github.com/openstax/kitchen/issues/284

### DIFF
--- a/lib/kitchen/directions/bake_chapter_glossary/v1.rb
+++ b/lib/kitchen/directions/bake_chapter_glossary/v1.rb
@@ -24,7 +24,8 @@ module Kitchen::Directions::BakeChapterGlossary
     def bake(chapter:, metadata_source:, append_to: nil, uuid_prefix: '', has_para: false)
       @glossary = []
 
-      # Use for books created by Adaptarr
+      # Use for books created by Adaptarr, where dd contains paragraphs
+      # More info: https://github.com/openstax/kitchen/issues/284
       if has_para
         chapter.glossaries.search('dd').each do |description|
           description_content = description.first('p').text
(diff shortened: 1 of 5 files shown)


lib/kitchen/directions/bake_chapter_glossary/v1.rb
```diff
@@ -21,9 +21,17 @@ def <=>(other)
       attr_reader :sortable
     end
 
-    def bake(chapter:, metadata_source:, append_to: nil, uuid_prefix: '')
+    def bake(chapter:, metadata_source:, append_to: nil, uuid_prefix: '', has_para: false)
       @glossary = []
 
+      # Use for books created by Adaptarr
```
brittany-johnson, line 27: Looks good - just one question to make sure I understand. 

This is fixing a transforms issue in some content created by Adaptarr and the team thought it would be best to resolve it in the recipe? 

Maybe it would be good to also link to the issue in this comment so in case someone needs further explanation as to where the paragraph is coming from they can refer to the issue? (if I'm understanding this correctly lol)

brittany-johnson (approved):

lib/kitchen/directions/bake_chapter_glossary/v1.rb
wilkus27: Sure :wink: I modified a comment.

In Adaptarr user have an ability to write description of the term in a few paragraphs (a few meanings). This is something that is not needed in Openstax books, but might be useful if we decide to use Adaptarr for another projects. So when you creating meaning of the term it puts your text automatically to paragraph. And this is something that we don't want to change in a tool.

PR #332, opened 2021-07-28
Add `ExampleElement#titles_to_rename` & refactor `BakeExample`

kswanson33:
`BakeExample` has been a problem because it targets titles to convert them to `h4`s, but examples can contain nested elements such as exercises, tables, and notes that shouldn't have their titles affected (except in the case of exercises? sometimes?). This approach would allow us to select `titles_to_rename` in an example & currently has a blacklist implementation. In the future, if other types of titles need to be skipped by BakeExample, this blacklist can be updated. We might find at some point that a whitelist implementation is better than a blacklist; in that case, the implementation can be changed.

### DIFF
--- a/lib/kitchen/directions/bake_example.rb
+++ b/lib/kitchen/directions/bake_example.rb
@@ -20,13 +20,7 @@ module Kitchen
                .pantry(name: :link_text)
                .store("#{I18n.t(:example_label)} #{number}", label: example.id)
 
-        example.titles.each do |title|
-          if title.parent.has_class?('os-caption-container') || \
-             title.parent.has_class?('os-caption') || \
-             title.parent.name == 'caption'
-            next
-          end
-
+        example.titles_to_rename.each do |title|
           title.name = 'h4'
         end
 
@@ -34,7 +28,6 @@ module Kitchen
           next if exercise.baked?
 
           if (problem = exercise.problem)
-            problem.titles.each { |title| title.name = 'h4' }
             problem.wrap_children(class: 'os-problem-container')
           end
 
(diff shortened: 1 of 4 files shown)


lib/kitchen/directions/bake_example.rb
```diff
@@ -20,21 +20,14 @@ def self.v1(example:, number:, title_tag:, numbered_solutions: false)
...
-
+        example.titles_to_rename.each do |title|
           title.name = 'h4'
         end
 
         example.exercises.each do |exercise|
           next if exercise.baked?
 
           if (problem = exercise.problem)
-            problem.titles.each { |title| title.name = 'h4' }
```
kswanson33, line 37: removed this because an exercise within an example should already have its title converted to an H4 by the above, & tables within exercises within examples were being mistakenly impacted

lib/kitchen/example_element.rb
```diff
@@ -27,9 +27,15 @@ def self.short_type
     #
     # @return [ElementEnumerator]
     #
-    def titles
-      search("[data-type='title']")
```
kswanson33, line 31: this can be removed because it's the same implementation as `ElementBase#titles`

lib/kitchen/example_element.rb
```diff
@@ -27,9 +27,15 @@ def self.short_type
...
+    def titles_to_rename
+      titles(except: \
+        lambda do |title|
+          title.parent.has_class?('os-caption-container') || \
+          title.parent.has_class?('os-caption') || \
+          title.parent.name == 'caption' || \
```
LostMojo, line 35: This will skip over uncooked captions, but I think that'll be okay?

LostMojo (approved):
Looks good! this will be v helpful

lib/kitchen/example_element.rb
kswanson33: yes, I think we probably don't want to impact tables when baking examples? Unless there's table-within-example-specific behavior, which I haven't seen yet

PR #340, opened 2021-08-02
EocCustomSectionToEocContainer refactor and EocSectionTitleLinkSnippet changes 

LostMojo:
Removes include_intro_page from arguments and fixes the numbering problems

### DIFF
--- a/spec/directions/bake_chapter_summary_spec.rb
+++ b/spec/directions/bake_chapter_summary_spec.rb
@@ -22,7 +22,7 @@ RSpec.describe Kitchen::Directions::BakeChapterSummary do
           <h1 data-type="document-title" itemprop="name" id="intro">Introduction page!</h1>
           <section class="summary" data-element-type="section-summary">
               <h3 data-type='title'>Kitchen Prep</h3>
-              <p>This should stay where it is.</p>
+              <p>This moves too!</p>
           </section>
         </div>
         <div data-type='page' id ="01">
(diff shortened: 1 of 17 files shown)


spec/directions/bake_chapter_summary_spec.rb
```diff
@@ -22,7 +22,7 @@
           <h1 data-type="document-title" itemprop="name" id="intro">Introduction page!</h1>
           <section class="summary" data-element-type="section-summary">
               <h3 data-type='title'>Kitchen Prep</h3>
-              <p>This should stay where it is.</p>
+              <p>This moves too!</p>
```
ajluk, line 25: @LostMojo I think it all looks good! Can I ask additionally about the lines from 46 to 62 below what are they need for?

ajluk (commented):
@LostMojo I just a question, because I don't remember why we need the non baking title in summary.

spec/directions/bake_chapter_summary_spec.rb
LostMojo: 46-62 checks that a title doesn't get baked twice if it's already been baked. [Karina added this behavior](https://github.com/openstax/kitchen/commit/bf6c0108edd4760f03be6d468e68183d69c21265#diff-c0bab3947e96d861b051f9a54b3f2876b5aba341e3a4359523ee4498ad77287e) a while back.

ajluk (approved):

PR #344, opened 2021-08-03
Non-introduction page count method

LostMojo:
Adds PageElement#count_in_chapter_without_intro_page and ChapterElement#has_introduction?

To better number pages when removed from their context

### DIFF
--- a/lib/kitchen/chapter_element.rb
+++ b/lib/kitchen/chapter_element.rb
@@ -44,6 +44,14 @@ module Kitchen
       pages('$.introduction').first
     end
 
+    # Returns true if the chapter has an introduction
+    #
+    # @return [Boolean]
+    #
+    def has_introduction?
+      introduction_page.present?
+    end
+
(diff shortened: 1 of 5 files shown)


lib/kitchen/chapter_element.rb
```diff
@@ -44,6 +44,14 @@ def introduction_page
       pages('$.introduction').first
     end
 
+    # Returns true if the chapter has an introduction
+    #
+    # @return [Boolean]
+    #
+    def has_introduction?
+      introduction_page.present?
```
jpslav, line 52: memoize this so that future calls are free.  

```ruby
def has_introduction?
  @has_introduction ||= introduction_page.present?
end
```

jpslav (changes requested):

lib/kitchen/chapter_element.rb
LostMojo: RIGHT

jpslav (approved):

PR #346, opened 2021-08-03
Add some answer key directions & refactor strategies

kswanson33:


lib/kitchen/directions/move_solutions_to_answer_key/move_solutions_from_exercise_section.rb
```diff
@@ -0,0 +1,40 @@
...
+    V1.new.bake(chapter: chapter, append_to: append_to, section_class: section_class,
+                title_number: title_number)
+  end
+
+  class V1
+    def bake(chapter:, append_to:, section_class:, title_number:)
+      solutions_clipboard = Kitchen::Clipboard.new
+      chapter.search("section.#{section_class}").exercises.each do |exercise|
+        solution = exercise.solution
+        next unless solution
```
LostMojo, line 14: This is very close to BakeChapterSolutions, but without the EOC page container. Should we condense some of the code there?

kswanson33: hmm... you're right that there's some shared functionality. They both cut solutions from exercise sections, found by section class, to a clipboard. The differences between them are a bit hairy though, because they extend to the method signature (BakeChapterSolutions takes an array of classes and this takes a single class). They also do different things with the solutions clipboard/append them to different elements. What do you think would be the best way to resolve this?

LostMojo (approved):

PR #350, opened 2021-08-05
BakeFirstElements select figures

LostMojo:
BakeFirstElements now select div.os-figure elements


KarinaMendez2:
looks good! mind adding the case in the spec? @LostMojo

kswanson33 (changes requested):
don't forget to add tests!

kswanson33 (approved):

PR #353, opened 2021-08-09
Expand Bake_Chapter_Introductions

LostMojo:
Added a new way to bake chapter objectives when they already exist in the introduction page in a div format

### DIFF
--- a/lib/kitchen/directions/bake_chapter_introductions.rb
+++ b/lib/kitchen/directions/bake_chapter_introductions.rb
@@ -7,22 +7,7 @@ module Kitchen
         book.chapters.each do |chapter|
           chapter_outline_html = ''
 
-          case bake_chapter_objectives
-          when 'div'
-            chapter_objectives = chapter.first('div.chapter-objectives')&.cut
-            if chapter_objectives
-              chapter_objectives.search('div[data-type="title"]')&.trash
-              chapter_objectives.wrap_children('div', class: 'os-note-body')
-              chapter_objectives.prepend(child:
-                <<~HTML
-                  <h3 class="os-title" data-type="title">
-                    <span class="os-title-label">#{I18n.t(:chapter_objectives)}</span>
-                  </h3>
-                HTML
-              )
-              chapter_outline_html = chapter_objectives.cut
-            end
-          when true
+          if bake_chapter_objectives
             outline_items_html = chapter.non_introduction_pages.map do |page|
               <<~HTML
                 <div class="os-chapter-objective">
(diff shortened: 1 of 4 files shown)


lib/kitchen/directions/bake_chapter_introductions.rb
```diff
@@ -7,9 +7,23 @@ def self.v1(book:, bake_chapter_objectives: true)
...
+              chapter_objectives.prepend(child:
+                <<~HTML
+                  <h3 class="os-title" data-type="title">
+                    <span class="os-title-label">#{I18n.t(:chapter_objectives)}</span>
+                  </h3>
+                HTML
+              )
+              chapter_outline_html = chapter_objectives.cut
+            end
+          when true
```
wilkus27, line 25: I'm not sure that we can use different types for method parameters :thinking: I think it can't be in one case boolean, and in another one string. I have two ideas:

1. We can use two parameters: `bake_chapter_outline` and `bake_chapter_objectives` of boolean type. Names are based on the final class for the element.
2. We can use `bake_chapter_objectives` parameter but it always be string type. What do you think about using strings: `modify` for the first case, because then chapter objectives exists in assembled file and we're just modifying it a little bit; and for second case it could be `add`, because we are creating new element, that is not exists in assembled file?

Let me know your thoughts :wink: Maybe we can also ask for an opinion second reviewer.

wilkus27 (changes requested):

lib/kitchen/directions/bake_chapter_introductions.rb
LostMojo: Hm, the first seems best, as these deal with two separate things (though I don't see them overlapping). It looks like I overcondensed the code a bit 😅

LostMojo: Oh because it would break the past recipes because they both deal with chapter objectives, that's why I went with the case. I think the different types for method parameters work here because of the bottom true/false at the end of the case statements. That end bit can catch whether the parameter exists or not, and won't crash anything. Maybe we should get a second opinion?

wilkus27: Oh I see :thinking: If we go with second solution all recipes have to be modified. And it's not what we want to do :smile: I asked @ajluk for an opinion and she said that she had similar problem, and it was resolved the same way as you did. Here is the link to PR: https://github.com/openstax/kitchen/pull/181#discussion_r644975380. There are also cases, and parameter type in method is boolean, but there is a case where it is Symbol.

So I think we can stay with current solution. I will give an approve :wink:

wilkus27 (approved):

lib/kitchen/directions/bake_chapter_introductions.rb
LostMojo: I actually talked with JP about this! I'm going to split this off into a v2 that uses symbols instead of strings to set the chapter_objective_strategy, and then depreciate the v1 method!

lib/kitchen/directions/bake_chapter_introductions.rb
```diff
@@ -60,6 +61,75 @@ def self.v1_update_selectors(something_with_selectors)
...
+              <<~HTML
+                <div class="os-chapter-objective">
+                  <a class="os-chapter-objective" href="##{page.title[:id]}">
+                    <span class="os-number">#{chapter.count_in(:book)}.#{page.count_in(:chapter)}</span>
+                    <span class="os-divider"> </span>
+                    <span data-type="" itemprop="" class="os-text">#{page.title.children[0].text}</span>
+                  </a>
+                </div>
+              HTML
+            end.join('')
```
wilkus27, line 95: I have a few questions :grin:

Second version is covering all cases. When `chapter_objective_strategy` is set to `default` it means that book has Chapter Outline - new element created from the beggining; when it's set to `preexisting_title` it has Chapter Objectives - element that exists in unbaked file and we modify it during baking, and it could be set to `nil` and then book doesn't have neither Chapter Outline, nor Chapter Objectives. Am I understanding correctly? If so, do we have a plan to remove first version in the future, when we update released recipes?  

I'm asking, because if we want to keep both versions, maybe second one should cover only case with Chapter Objectives? So if book has Chapter Outline recipe should use V1, when it has Chapter Objectives - V2.

Also, if we want to have two versions, maybe they should be split into separate files, to keep the convension. But it 's a small thing :wink:

To sum up, for the first scenario (removing V1 in the future) maybe good solution will be to add comment that V1 is used only in released books, and all new recipes should use V2. 

For the second one (keeping both versions) as I write before, we can split code into two files and remove part for Chapter Outline from V2.

Let me know what do you think, and wheter I understand everything correctly. I might miss something :slightly_smiling_face:

wilkus27: I'm sorry, I just noticed that I missed last part of your sentence, that you are going to depreciate the v1 method :facepalm: Sorry again :grin: Will that be done in this PR or in the future? If in the future, I think the comment which I mentioned before might be helpful. Beside that everything looks good :relaxed:

LostMojo: For now we're leaving V1 _mostly_ alone because it does some chapter_outline stuff that we stumbled adding to a V2, but I think we could eventually depreciate V1

LostMojo:
Closed because the issues are being added to Karina's issue: https://github.com/openstax/kitchen/pull/356

Issue #349, opened 2021-08-04
Create BakeAnnotationClasses direction for English Composition


ajluk:
Paragraphs with `annotation` classes  present in English Composition need to be supported in Kitchen. Corresponding `recipes` PR: https://github.com/openstax/recipes/pull/69
Screenshot showing differences between cnx-recipes and kitchen baked output file for such paragraphs: 
![image](https://user-images.githubusercontent.com/55203598/128886053-1c4b6668-f780-4590-90ae-a56903940e20.png)

ajluk:
Related `cnx-recipes` issue that needs to be checked out in Kitchen: https://github.com/openstax/cnx-recipes/issues/4015

PR #354, opened 2021-08-09
create bakeannotationclasses direction

ajluk:
Issue: #349
It creates `BakeAnnotationClasses` directions for baking paragraphs with annotation classes where the inner content from such para is moved to separate `span` inside it and a div container containing icons `spans` with corresponding classes presented in paragraphs is created.
Pr containing similar changes for English Composition from `cnx-recipes`: https://github.com/openstax/cnx-recipes/pull/3617
Corresponding `recipes` PR with English Composition kitchen recipe: https://github.com/openstax/recipes/pull/69

### DIFF
--- a/lib/kitchen/directions/bake_annotation_classes/v1.rb
+++ b/lib/kitchen/directions/bake_annotation_classes/v1.rb
@@ -19,7 +19,7 @@ module Kitchen::Directions::BakeAnnotationClasses
                                    auditory-icon kinesthetic-icon]
       annotation_icon_classes.each do |annotation_icon_class|
         chapter.search("p.#{annotation_icon_class}").each do |annotation_with_icon_class|
-          annotation_with_icon_class.search('.os-icons').first.append(child:
+          annotation_with_icon_class.search('div.os-icons').first.append(child:
             <<~HTML
               <span class = "#{annotation_icon_class}"></span>
             HTML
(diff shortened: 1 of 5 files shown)


lib/kitchen/directions/bake_annotation_classes/v1.rb
```diff
@@ -0,0 +1,31 @@
...
+            <div class="os-icons"></div>
+          HTML
+        )
+      end
+      annotation_icon_classes = %w[linguistic-icon
+                                   culture-icon
+                                   dreaming-icon
+                                   visual-icon
+                                   speech-icon
+                                   auditory-icon kinesthetic-icon]
```
KarinaMendez2, line 19: oop, new line for last class

lib/kitchen/directions/bake_annotation_classes/v1.rb
```diff
@@ -0,0 +1,31 @@
...
+      annotation_icon_classes.each do |annotation_icon_class|
+        chapter.search("p.#{annotation_icon_class}").each do |annotation_with_icon_class|
```
KarinaMendez2, line 21: try this new feature! 

`chapter.search(annotation_icon_class.prefix('p.').join(', ')).each do |annotation_with_icon_class|`

KarinaMendez2 (changes requested):

lib/kitchen/directions/bake_annotation_classes/v1.rb
ajluk: @KarinaMendez2 I tried to refactor the lines into one loop using the `prefix` feature, but got stuck when there are one that one icon classes in a paragraph which need to be places in separate `spans`: 
image
image
I guess in this case we must keep the outer loop. If you will have some other idea please let me know :slightly_smiling_face:

KarinaMendez2 (approved):

PR #355, opened 2021-08-10
add ChangeSubsectionTitleTag direction

kswanson33:
so far needed in contemporary math, precalc, and college algebra


ajluk (approved):
@kswanson33 I'm not sure why the checks are not passing, but I'm approving because it all looks good here :slightly_smiling_face:

PR #356, opened 2021-08-11
Bake chapter intro v2

KarinaMendez2:


spec/directions/bake_chapter_introductions_spec.rb
```diff
@@ -179,4 +234,120 @@
...
+                    <p id="123_copy_1">This chapter begins with an overview of anatomy and...</p>
+                  </div>
+                </div>
+              </div>
+            </div>
+          </body>
+        HTML
+      )
+    end
+  end
```
wilkus27, line 273: @KarinaMendez2 In which book we can find this case? I think it is a mistake in old recipes. All notes should be modified like in case from below, so content should be wrapped into `os-note-body` and title should be modified to h3. I guess that someone didn't add note with class `chapter-objectives` to `Config_Note`. It should be done like that in cnx-recipes:

image

lib/kitchen/directions/bake_chapter_introductions/v1.rb
```diff
@@ -0,0 +1,54 @@
...
+          end
+
+        final_section =
+          if bake_chapter_outline && bake_chapter_objectives
+            Kitchen::Directions::BakeChapterIntroductions.bake_chapter_outline(
+              chapter_objectives_html: chapter_objectives_html
+            )
+          else
+            chapter_objectives_html
+          end
```
wilkus27, line 36: I'm not sure if I understand this part correctly :thinking: Why do we need `bake_chapter_objectives:` parameter? First version in only for Chapter Outline, that's right? It exists or not. And if it exists it always contain default Chapter Objectives (defined in `BakeChapterObjectives`). 

I'm wondering if something like that could work:

```
if bake_chapter_outline
  chapter_objectives_html = Kitchen::Directions::BakeChapterIntroductions.bake_chapter_objectives(
              chapter: chapter
            )
  chapter_outline = Kitchen::Directions::BakeChapterIntroductions.bake_chapter_outline(
              chapter_objectives_html: chapter_objectives_html
            )
end
```
and in the end:

```
introduction_page.append(child:
          <<~HTML
            <div class="intro-body">
              #{chapter_outline if bake_chapter_outline}
              <div class="intro-text">
                #{title.paste}
                #{intro_content.paste}
              </div>
            </div>
          HTML
        )
```

Did I miss any case?

lib/kitchen/directions/bake_chapter_introductions/v2.rb
```diff
@@ -0,0 +1,44 @@
...
+        title.name = 'h2'
+        Kitchen::Directions::MoveTitleTextIntoSpan.v1(title: title)
+
+        chapter_objectives_html =
+          Kitchen::Directions::BakeChapterIntroductions.bake_chapter_objectives(
+            chapter: chapter,
+            chapter_objectives_strategy: chapter_objectives_strategy
+          )
+
+        intro_content = introduction_page.search('[data-type="note"].chapter-objectives').cut
```
wilkus27, line 22: As I write in my other comment I think there is a mistake in old recipe. Probably this part is not necessery. Note should be modified like it is done for `chapter_objectives_html`.

spec/directions/bake_chapter_introductions_spec.rb
KarinaMendez2: This case of not having chapter-outline is Anatomy & P, first case I run into

lib/kitchen/directions/bake_chapter_introductions/v1.rb
KarinaMendez2: I talked to Alina and she says that there are book that do have one or the other, so we might run into books needing chapter objectives but not chapter outline

KarinaMendez2:
@wilkus27 @LostMojo  Talked to Alina, at the moment, the safest decision is to leave both versions (old way of baking chapter-objectives and the way: treat them as a note) so that we don't break other books and when Larissa comes back we can re-visit.

lib/kitchen/directions/bake_chapter_introductions/v1.rb
wilkus27: Yes, but in situation when there is Chapter Objectives but there is no Chapter Outline we are using V2. Do we need this in V1? I guess all previous (released) books have Chapter Outline. First book that has only Chapter Objectives was one that Mary Jo was working on and now you have another book, that's right? If yes, then maybe we can remove this case from V1 :thinking:

lib/kitchen/directions/bake_chapter_introductions/bake_chapter_objectives.rb
```diff
@@ -0,0 +1,44 @@
+# frozen_string_literal: true
+
+module Kitchen::Directions::BakeChapterIntroductions
+  class BakeChapterObjectives
+    def bake(chapter:, chapter_objectives_strategy:)
+      chapter_outline_html = ''
```
wilkus27, line 6: What do you think to rename `chapter_outline_html` to `chapter_objectives_html` and `chapter_objectives` to `chapter_objectives_note` or something like that?

image (8)

Later in `BakeChapterOutline` we are using this name, so maybe we can keep it consisntent.

image (9)

It's just a suggestion :wink: Let me know your thoughts.

lib/kitchen/directions/bake_chapter_introductions/v2.rb
wilkus27: Do we have a decision from Larissa about that?

wilkus27:
@KarinaMendez2 I added one comment about renaming variables. It's just a proposition, so I didn't label it as a requested change :wink: 

I have a question about Chapter Objectives from Anatomy. Do we have a decision from Larissa what to do with that? Do we want to keep it like it is in old recipe (so just copy and paste into intro body) or do we want to fix it, so modify like it is done in Mary Jo book (so to use `:preexisting_title` strategy)?

Also I think we should clean up specs. Now we have two versions of baking Introduction Page, so probably we should create separate files with specs for each version. What do you think?

KarinaMendez2:
@wilkus27  Hi Natalia! Talked with Larissa yesterday and she said yes, to treat chapter objectives as notes (preexisting_title strategy) so I started working on it yesterday and will continue today, I will let you know when it's ready for review again, thank you for your suggestions!

lib/kitchen/directions/bake_chapter_introductions/bake_chapter_objectives.rb
```diff
@@ -0,0 +1,26 @@
...
+
+      case chapter_objectives_strategy
+      when :default
+        # trash existing title
+        chapter.notes('$.chapter-objectives').titles.first&.trash
+        Kitchen::Directions::BakeAutotitledNotes.v1(
+          book: chapter,
+          classes: %w[chapter-objectives],
+          bake_subtitle: false
+        )
```
LostMojo, line 16: Nice!

lib/kitchen/directions/bake_chapter_introductions/v2.rb
```diff
@@ -0,0 +1,44 @@
...
+                #{title.paste}
+                #{extra_content.paste}
+              </div>
+            </div>
+          HTML
+        )
+      end
+
+      Kitchen::Directions::BakeChapterIntroductions.v1_update_selectors(book)
+    end
```
LostMojo, line 42: So V2 has no way to bake chapter outlines? Am I reading that right?

LostMojo (commented):
I need a bit of clarification on one part, but the parts that effect my book look good!

lib/kitchen/directions/bake_chapter_introductions/bake_chapter_objectives.rb
```diff
@@ -0,0 +1,44 @@
...
+        end.join('')
+      when :preexisting_title
+        chapter_objectives = chapter.first('[data-type="note"].chapter-objectives')&.cut
+        if chapter_objectives
+          chapter_objectives.search('div[data-type="title"]')&.trash
+          chapter_objectives.wrap_children('div', class: 'os-note-body')
+          chapter_objectives.prepend(child:
+            <<~HTML
+              <h3 class="os-title" data-type="title">
+                <span class="os-title-label">#{I18n.t(:chapter_objectives)}</span>
```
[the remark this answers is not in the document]
LostMojo: Oh god it published all the ones that never got published in the first round of review I did 😓

lib/kitchen/directions/bake_chapter_introductions/v2.rb
KarinaMendez2: oop, since we ran into books that don't need it, I didn't include it in v2, I was thinking that when book recipes get refactored and start using v2, we could add another strategy to call `bake_chapter_outline`, since the books that do bake a chapter outline are using v1. Do you think it's necessary to do it in this PR?

LostMojo: If it's depreciating v1, then v2 needs to cover the functionality of v1 :T 

If it's quick, you could add that in, but we just want to move forward, we could un-depreciate v1, and then make a new card to make v2 have a strategy for baking chapter outlines as well

LostMojo (approved):

PR #359, opened 2021-08-19
Create direction for baking injected exercises

kswanson33:


lib/kitchen/element_base.rb
```diff
@@ -772,6 +772,15 @@ def as_enumerator
...
 
+    # Returns an enumerator over exercises of all types
+    # dynamically cast to ExerciseElement or InjectedQuestionElement at runtime
+    #
+    # @return [TypeCastingElementEnumerator]
+    #
+    def all_exercise_types
+      search_with(Kitchen::ExerciseElementEnumerator, Kitchen::InjectedQuestionElementEnumerator)
+    end
+
```
kswanson33, line 783: @jpslav can you confirm that this is a good way to do this & good place to put this 😃

jpslav (changes requested):

PR #380, opened 2021-09-09
Fix BakeStepwise to skip nested lists

kswanson33:


ajluk (approved):
All looks good! :slightly_smiling_face:

PR #385, opened 2021-09-15
Learning Objectives sections

wilkus27:
In BCA Learning Objectives is a section with class `learning-objectives` and existing title, no div with `data-type="abstract"` created from metadata. Only thing that should be done during baking is numbering.


wilkus27:
Done in https://github.com/openstax/kitchen/pull/387. I'm closing :smile:

PR #389, opened 2021-09-16
 BakeAutotitledExercise 

KarinaMendez2:


lib/locales/en.yml
```diff
@@ -41,3 +41,7 @@ en:
     access_for_free: Access for free at openstax.org
   index:
     main: Index
+  exercises:
+    your-turn: Your Turn
```
LostMojo, line 45: Shouldn't this be recipe side if it's book-specific?

LostMojo (changes requested):

lib/locales/en.yml
KarinaMendez2: Probably 🤔 i’ll move it into my recipe yml

wilkus27 (approved):
Looks good :wink: :+1:

LostMojo (approved):

PR #390, opened 2021-09-17
Add numbering to references

KarinaMendez2:
for: https://app.zenhub.com/workspaces/ce-styles-5cdc84baee941c07853fbbf2/issues/openstax/recipes/94

### DIFF
--- a/lib/kitchen/directions/bake_references/v1.rb
+++ b/lib/kitchen/directions/bake_references/v1.rb
@@ -28,9 +28,12 @@ module Kitchen::Directions::BakeReferences
         end
 
         chapter_references = chapter.pages.references.cut
-        chapter_title = chapter.title.search('.os-text')
 
-        chapter_title = chapter.title.search('.os-number, .os-divider, .os-text') if numbered_title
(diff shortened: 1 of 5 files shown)


lib/kitchen/directions/bake_references/v1.rb
```diff
@@ -28,12 +28,14 @@ def bake(book:, metadata_source:)
         end
 
         chapter_references = chapter.pages.references.cut
-        chapter_title_no_num = chapter.title.search('.os-text')
+        chapter_title = chapter.title.search('.os-text')
+
+        chapter_title = chapter.title.search('.os-number, .os-divider, .os-text') if numbered_title
```
wilkus27, line 33: I'm wondering if it's ok to have two definitions :thinking: What do you think to use `if else`?

```
chapter_title = if numbered_title do
                          chapter.title.search('.os-number, .os-divider, .os-text')
                       else
                          chapter.title.search('.os-text')
                       end
```

wilkus27 (changes requested):

wilkus27 (approved):

PR #402, opened 2021-09-28
add bake unnumbered figures in preface

ajluk:
In World History, there are unnumbered figures present in Preface, instead adding another loop searching preface in recipe I added baking such figures in preface by default, because there could be more cases in the old books.

### DIFF
--- a/lib/kitchen/directions/bake_figure.rb
+++ b/lib/kitchen/directions/bake_figure.rb
@@ -3,7 +3,7 @@
 module Kitchen
   module Directions
     module BakeFigure
-      def self.v1(figure:, number:, cases: false)
+      def self.v1(figure:, number: nil, cases: false)
         return if figure.has_class?('unnumbered') && !figure.has_class?('splash') && !figure.caption
 
         figure.wrap(%(<div class="os-figure#{' has-splash' if figure.has_class?('splash')}">))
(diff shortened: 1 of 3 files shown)


lib/kitchen/figure_element.rb
```diff
@@ -47,16 +47,42 @@ def subfigure?
...
     #
     # @return [Boolean]
-    #
+
+    # TODO, remove this method from other books recipes to prevent counting unnumbered baked figures
     def figure_to_bake?
-      return false if subfigure? || (has_class?('unnumbered') && !has_class?('splash') && !caption)
+      return false if subfigure? || (has_class?('unnumbered') &&
+                                    !has_class?('splash') && !caption &&
+                                    !title)
```
ajluk, line 68: I have updated `figure_to_bake` method to contain a title and left the method in place to avoid breaking the other recipes. Below I have created two separate methods for baking numbered and unnumbered figures (`figure_to_bake_without_count`, `figure_to_bake_and_count` ). The reason for it was that when I tried to use `figure_to_bake` in WH recipe I couldn't exclude unnumbered figures from counting while looping through all figures that should be baked: 
image

also attempts to define the figures to be numbered under condition gave no results: 

image
image
In various books amount of figures can differ and in some of the them using `figure_to_bake` could not break the numbering (f.ex. if there will be no unnumbered figure in a chapter).

My second thought was to create two separate directions for baking unnumbered and numbered figures. If you think it will be a better idea or you will have some other suggestions please let me know :relaxed:

PR #410, opened 2021-10-04
add SectionElement and SectionElementEnumerator

kswanson33:


ajluk (approved):
Everything looks good! :slightly_smiling_face:

ajluk (changes requested):
Kendra I have already approved, but then I realized that the specs are missing. :see_no_evil:

ajluk (approved):

PR #422, opened 2021-10-13
Fix index for Polish books

wilkus27:
Issue: https://github.com/openstax/kitchen/issues/442


ajluk (approved):

lib/kitchen/patches/i18n.rb
```diff
@@ -4,6 +4,10 @@
 
 # rubocop:disable Style/Documentation
 module I18n
+  def self.group_by(character)
```
kswanson33, line 7: `group_by` as a name makes sense in the context of indexes, but this function doesn't group anything, it just decides what character to return based on locale. Maybe a better name would be something like `transliterate_for_index` or `character_to_group`, or something I haven't thought of! lmk what you think :)

spec/helpers/stub_helpers.rb
```diff
@@ -1,17 +1,19 @@
...
     I18n.config.available_locales = %i[test en es pl]
     allow_any_instance_of(I18n::Config).to receive(:backend).and_return(
       I18n::Backend::Simple.new.tap do |backend|
-        backend.store_translations 'test', hash
+        backend.store_translations language, hash
       end
     )
 
-    assign_i18n_dot_locale(:test)
+    assign_i18n_dot_locale(language)
```
kswanson33, line 16: nice

kswanson33 (changes requested):
I like what you've done here! One naming-related comment, but otherwise it looks great to me 🎉

lib/kitchen/patches/i18n.rb
wilkus27: Oh good point :+1: I didn't think about it. I like very much second proposition. It's more descriptive. I made a fix for that.

Issue #427, opened 2021-10-25
Links in Exercises (Accounting vol1) need chapter content pages target labels 


ajluk:
In all of the exercises in Accounting vol1 there are links that should target corresponding chapter content pages and therefore there is a need to create target labels for these pages in Kitchen optionally:
![image](https://user-images.githubusercontent.com/55203598/138728561-cd538d5a-7a70-434b-864f-beaab8175802.png)
In this case also spans `.label-text` , `.label-counter` are needed due to webview requirements: 
![image](https://user-images.githubusercontent.com/55203598/140356345-26a9322c-fea3-4d42-990a-9961835ad980.png)

PR #428, opened 2021-10-25
Pages target labels accountin exercises

ajluk:
Issue:  #427


lib/kitchen/directions/bake_link_placeholders.rb
```diff
@@ -26,6 +26,14 @@ def self.v1(book:, cases: false)
...
           if replacement.present?
             anchor.replace_children(with: replacement)
+            if label_wrappers && anchor.has_class?('lo-reference')
+              anchor.wrap_children('span', class: 'label-counter')
+              anchor.prepend(child:
+                <<~HTML
+                  <span class="label-text">#{I18n.t(:lo_label_text)}</span>
+                HTML
+              )
+            end
```
kswanson33, line 36: This seems to be a pretty specific case. The `label_wrappers` argument doesn't immediately make it clear that this argument is basically just to add the `LO` label to learning objectives links. 

Instead of doing this within `BakeLinkPlaceholders`, I think a separate `BakeLOLinkLabel` direction could handle this. That way, it's easier to remember what the direction is for, and keeps it from getting buried in `BakeLinkPlaceholders`.

kswanson33 (changes requested):
This is a complicated one! I like the approach in `BakeNonIntroPages`. Let me know what you think about my suggestion re: `BakeLinkPlaceholders`

kswanson33 (approved):
woo! looks good 🎉

PR #432, opened 2021-10-27
add optional figure reference baking injected exercises

ajluk:
Issue: https://github.com/openstax/cnx-recipes/issues/4177


lib/kitchen/directions/bake_injected_exercise/bake_injected_exercise_question.rb
```diff
@@ -41,11 +42,15 @@ def bake(question:, number:, only_number_solution:)
         end
       end
 
+      context = question.ancestor(:injected_exercise).exercise_context&.cut&.paste if figure_reference
```
ajluk, line 45: @kswanson33 I have tried refactored the code with your requested changes but I've got stuck in this particular place and got such error: 
image
In my recipe I'm using this loop, where it is not pointing to `injected_exercises but `exercise_question``: 
image
and I guess this is the reason why I can't select an ancestor for the `exercise_question` of given type (`:injected_exercise`). Do you think there is a need to change the `BakeInjectedExerciseQuestion`  to point `injected_exercises` instead of  `exercise_question`? Or maybe you can see a different approach here, rewrite the loop in the recipe somehow? Let me know if you will have some suggestions.

ajluk: @kswanson33 could you also explain what did you mean by `single-part exercises`?

ajluk: Small EDIT:
I have also tried to use the element class `InjectedExerciseElement` itsef: 
image
but also got error: 
image

ajluk:
Hi @kswanson33 After Larissa's comment I have reworked the PR to only move the `exercise-context` if there is only one `exercise-question` per `injected-exercise` so this feature will be for now available only for singular part exercises. I have also confirmed it with Larissa. I was also thinking if there will be more exercises questions when there is a `exercise-context` figure present in some books we perhaps should tag the exercise question with some class, so we will know where to move the `exercise-context` figure reference but Larissa needs to look into it deeper how the content will look.

CHANGELOG.md
```diff
@@ -8,7 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 * Add `BakeScreenreaderSpans` direction (minor)
 * Fix `BakeIndex` to group terms by character in polish books and transliterate it for others (minor)
-
+* Add `figure_reference` option for `BakeNoteInjectedQuestion` to bake exercise-context figure reference (minor)
```
kswanson33, line 11: Don't forget to update the changelog to reflect the new change! I believe this change is `major` as single-part injected exercises should always be changed.

lib/kitchen/directions/bake_injected_exercise/bake_injected_exercise_question.rb
```diff
@@ -1,12 +1,13 @@
...
-  def self.v1(question:, number:, only_number_solution: false)
-    V1.new.bake(question: question, number: number, only_number_solution: only_number_solution)
+  def self.v1(question:, number:, only_number_solution: false, figure_reference: false)
+    V1.new.bake(question: question, number: number, only_number_solution: only_number_solution,
+                figure_reference: figure_reference)
   end
 
   class V1
-    def bake(question:, number:, only_number_solution:)
+    def bake(question:, number:, only_number_solution:, figure_reference:)
```
kswanson33, line 10: do we need an argument here? probably this should always be true

kswanson33 (changes requested):
looks really good! two little comments but other than that it's fantastic :)

kswanson33 (approved):

PR #437, opened 2021-11-03
BakeNumberedExercises Appendix Exercise rules added

LostMojo:
It now bakes exercises in appendices!

### DIFF
--- a/lib/kitchen/directions/bake_numbered_exercise/v1.rb
+++ b/lib/kitchen/directions/bake_numbered_exercise/v1.rb
@@ -8,7 +8,7 @@ module Kitchen::Directions::BakeNumberedExercise
       problem = exercise.problem
       solution = exercise.solution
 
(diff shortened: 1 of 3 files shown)


lib/kitchen/directions/bake_numbered_exercise/v1.rb
```diff
@@ -61,6 +78,18 @@ def bake_solution(exercise:, number:, solution_stays_put:, divider: '. ')
         return
       end
 
+      if in_appendix
+        solution.id = "#{exercise.id}-solution"
```
kswanson33, line 82: are lines 82-83 redundant with lines 93-94?

kswanson33: or, could lines 93-94 go above the `in_appendix` if-statement since they happen in both cases?

kswanson33 (changes requested):
Looks pretty good, just some little things!

kswanson33 (approved):

PR #446, opened 2021-11-17
Add #preceded_by_text method to Node_Element and Nokigiri

LostMojo:
Figures and tables need a #previous method that returns the previous node (while not returning empty spaces or text not wrapped in a node)

[Anna's reference PR ](https://github.com/openstax/kitchen/pull/443/files) needs a #previous method that does not return empty spaces (like newlines or tabs) but DOES return unwrapped text if it occurs immediately previous. (It's checking whether or not a comma needs to go in between similar elements)

Edit: Changed to add a #preceded_by_text? method in node_element and nokigiri.

Edit2: We might not need the nokigiri patched #previous method?

### DIFF
--- a/lib/kitchen/element_base.rb
+++ b/lib/kitchen/element_base.rb
@@ -519,6 +519,13 @@ module Kitchen
       )
     end
 
+    # returns boolean
+    # depending on whether there is non-contained text right before the element
+    #
+    def preceded_by_text?
+      raw.preceded_by_text?
+    end
+
     # TODO: make it clear if all of these methods take Element, Node, or String
 
     # If child argument given, prepends it before the element's current children.
--- a/lib/kitchen/element_base.rb
+++ b/lib/kitchen/element_base.rb
@@ -82,9 +82,13 @@ module Kitchen
     #   Set the inner HTML for this element
     #   @see https://www.rubydoc.info/github/sparklemotion/nokogiri/Nokogiri/XML/Node#inner_html=-instance_method Nokogiri::XML::Node#inner_html=
     #   @return Object
+    # @!method preceded_by_text
+    #   Returns true if the immediately preceding sibling is text
+    #   @return Boolean
     def_delegators :@node, :name=, :name, :[], :[]=, :add_class, :remove_class,
                    :text, :wrap, :children, :to_html, :remove_attribute,
-                   :key?, :classes, :path, :inner_html=, :add_previous_sibling
+                   :key?, :classes, :path, :inner_html=, :add_previous_sibling,
+                   :preceded_by_text?
 
     # @!method config
     #   Get the config for this element's document
@@ -515,13 +519,6 @@ module Kitchen
         document: document,
         short_type: "previous(#{short_type})"
       )
-    end
-
-    # returns boolean
-    # depending on whether there is non-contained text right before the element
-    #
-    def preceded_by_text?
-      raw.preceded_by_text?
     end
 
     # TODO: make it clear if all of these methods take Element, Node, or String
--- a/lib/kitchen/element_base.rb
+++ b/lib/kitchen/element_base.rb
@@ -507,7 +507,7 @@ module Kitchen
     # nil if there's no previous sibling
     #
     def previous
-      prev = raw.previous
+      prev = raw.previous_element
       return prev if prev.nil?
 
       Element.new(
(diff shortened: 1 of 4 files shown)


lib/kitchen/patches/nokogiri.rb
```diff
@@ -71,7 +71,15 @@ def previous
         prev = previous_element
```
LostMojo, line 71: So this is actually a different method than previous/previous_sibling that only returns the previous element. We might not need the updated previous method or text checks at all.

lib/kitchen/element_base.rb
```diff
@@ -504,6 +504,7 @@ def parent
 
     # returns previous element
     # skips double indentations that the nokigiri sometimes picks up
+    # also skips text belonging to the parent node
     # nil if there's no previous sibling
     #
     def previous
```
LostMojo, line 510: Switch the nokigiri#previous call to a nokigiri#previous_element?

spec/element_base_spec.rb
```diff
@@ -134,11 +140,33 @@
       )
     end
 
+    it 'skips text belonging to a parent element' do
+      expect(reference.previous).to match_normalized_html(
+        <<~HTML
+          <div data-type="note">Reference 1</div>
+        HTML
+      )
+    end
```
kswanson33, line 149: maybe instead of "skips text belonging to a parent" it's more accurate to say that this "skips sibling text nodes" or "only returns element sibling"

lib/kitchen/element_base.rb
```diff
@@ -504,6 +504,7 @@ def parent
 
     # returns previous element
     # skips double indentations that the nokigiri sometimes picks up
+    # also skips text belonging to the parent node
```
kswanson33, line 507: same comment here, this is a bit confusing

lib/kitchen/patches/nokogiri.rb
```diff
@@ -74,6 +74,14 @@ def previous
         prev.text? ? prev.previous : prev
       end
 
+      def preceded_by_text?
+        prev = previous_sibling
+        while !prev.nil? && prev.blank? do prev = prev.previous_sibling end
```
kswanson33, line 79: this is a good way of getting through a bunch of blank nodes! it doesn't seem likely that the while loop could end up looping infinitely, but it might be a good thing to check with a few more test cases.

kswanson33 (commented):
interested in what is this for -- is it for a direction? good use of the patch architecture, this looks great overall

lib/kitchen/patches/nokogiri.rb
```diff
@@ -74,6 +74,14 @@ def previous
...
+        return false if prev.nil?
+
+        prev.text?
+      end
```
jpslav, line 83: Why is this using `previous_sibling` whereas `previous` above is using `previous_element`?  

I'm also confused now why the `previous` implementation even checks for `text?` because per the [Nokogiri docs](https://www.rubydoc.info/github/sparklemotion/nokogiri/Nokogiri/XML/Node#previous_element-instance_method), `previous_element` only returns element node types, i.e. not text nodes???

lib/kitchen/element_base.rb
LostMojo: Actually, it _would_ skip text, or at least was doing so when I tried to use it in the #preceded_by_text method. previous_element only returns the next previous element-- no text. I'm not sure why this behavior was missed the first time around, but here we are.

lib/kitchen/patches/nokogiri.rb
LostMojo: I'm not entirely sure--- it's using previous_sibling to pick up text-- as I said in the previous comment, apparently previous_element on paper does exactly what we want it to do-- (instead of the  nokigiri #previous/#previous sibling which just returns the previous node) perhaps we could get rid of the #previous patch, and switch Node_Element#previous to use raw.previous_element?

LostMojo:
Maybe this should go in a different PR-- but I've removed the nokigiri#previous patch (keeping the note_element#previous intact, it just uses #previous_element now) and none of the specs fail, including ones I've written to specifically test for the blank space bug.

In kitchen, the only 'previous' method used is the note_element#previous, so I don't see any potential problems arising from this change.

LostMojo:
Okay, split off the thoughts regarding #previous / #previous_element redundancy into a separate PR #449.

Basically: I think nokigiri#previous_element has all the functionality we wanted from nokigiri#previous, and switching NodeElement#previous to depend on nokigiri#previous_element doesn't cause any specs to fail (and only NodeElement#previous is called elsewhere in kitchen. not nokigiri#previous)

kswanson33 (approved):

PR #457, opened 2021-11-26
answer key containers class

ajluk:
Done for accounting recipe.


ajluk:
Closing since the class will be changed to `os-solutions-container` in the recipe to be consistent with the rest of the books in Kitchen.

PR #462, opened 2021-12-02
add method that generate roman numbers up to 100

wilkus27:
Done for Intellectual Property. This book has over 80 footnotes in roman numbers, so I added method that generate them rather then create bigger constant. I had problems with instance variable. Adding `attr_reader` or `attr_accesor` at the beginning of the class didn't help, methods from this class treated variable as nil. I found another solution, but if someone have better idea how to handle this let me know :wink:


LostMojo (approved):
It looks good to me! I'd maybe add a spec to test if it gets 9s correctly, but otherwise, it looks pretty nice!

PR #465, opened 2021-12-13
fix searching cases in target labels

wilkus27:
Done for Polish U-Physics.

After moving sections to EOC target labels are loosing `cmlnle` part of attribute.


KarinaMendez2 (approved):

wilkus27:
Same problems in Index. Probably cutting to clipboard is removing namespaces prefixes. I couldn't find why it acts like that. That's why I decided to add just additional selectors without prefixes to methods.

KarinaMendez2 (approved):

PR #471, opened 2021-12-21
add unit title prefix number unit page

ajluk:
For: https://github.com/openstax/cnx-recipes/issues/4316

### DIFF
--- a/spec/directions/bake_toc_spec.rb
+++ b/spec/directions/bake_toc_spec.rb
@@ -29,6 +29,9 @@ RSpec.describe Kitchen::Directions::BakeToc do
           </h1>
           <div data-type="page" id="p2">
             <h2 data-type="document-title">
+              <span class="os-part-text">Unit </span>
+              <span class="os-number">1</span>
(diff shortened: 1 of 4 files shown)


spec/directions/bake_toc_spec.rb
```diff
@@ -344,6 +347,8 @@
               <ol class="os-unit">
                 <li class="os-toc-unit-page" cnx-archive-shortid="" cnx-archive-uri="p2">
                   <a href="#p2">
+                    <span class="os-number"><span class="os-part-text">Unit </span>1</span>
```
LostMojo, line 350: is the os-part-text supposed to be inside the os-number span like that? It seems a little unusual

LostMojo (approved):

spec/directions/bake_toc_spec.rb
ajluk: @LostMojo as we discussed on slack this part of toc was already there originally created by JP. :slightly_smiling_face:

PR #480, opened 2022-01-28
v19.0.0

LostMojo:


Gemfile.lock
```diff
@@ -27,7 +27,7 @@ GEM
     concurrent-ruby (1.1.9)
     diff-lcs (1.3)
     docile (1.4.0)
-    i18n (1.8.11)
+    i18n (1.9.1)
```
ajluk, line 30: @LostMojo Yesterday @kswanson33 has fixed the gem version bug in recipes by adding a line with i18n gem installation to `install_used_gem_ version` file in `recipes` repo, I'm wondering if this should be also updated there now as well :thinking:

ajluk (approved):

